PR #15, opened 2019-11-29
small fix for LOCAL_SECRET_JSON

ch-stark:
without the fix only pull-secret.json would be valid input


schmaustech:
This lgtm.

ch-stark:
after further discussion this PR is not needed and will be resolved in a different way

PR #17, opened 2019-11-29
Fix local pull secret issue

schmaustech:


ch-stark:
a first test worked fine for me

PR #19, opened 2019-12-01
Add instructions how to check huge pages

cynepco3hahue:
Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/features/performance/README.md
+++ b/features/performance/README.md
@@ -1,0 +1,19 @@
+# Performance Tuning
+
+## Huge Pages
+
+To verify huge pages feature functionality:
+
+- label your worker nodes with `oc label node <node_name> machineconfiguration.openshift.io/role=worker-rt`
+- apply huge pages kernel boot parameters config via `oc create -f manifests/05-mc-kargs-worker-rt.yaml`
+- wait for workers update
+
+```bash
+oc -n openshift-machine-config-operator wait machineconfigpools worker --for condition=Updating --timeout=1800s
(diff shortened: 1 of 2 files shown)


features/hugepages/examples/hugepages-machine-config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+    labels:
+        machineconfiguration.openshift.io/role: hugepages
+    name: hugepages
+spec:
+    kernelArguments:
```
e-minguez: Just in case, I'd say the kernelArguments would be better if they are quoted (as in https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfiguration.md#kernelarguments )

cynepco3hahue: hm I think in general it does not really matter until you have some special characters under the value, but I do not have problems to add quotes under the PR.

cynepco3hahue: done

simon3z:
@cynepco3hahue discuss with @yanirq @slintes @vladikr where to keep the single MachineConfig.
cc @MarSik

features/performance/README.md
```diff
@@ -0,0 +1,19 @@
+# Performance Tunning
```
slintes: nit: Tuning

cynepco3hahue: done

PR #20, opened 2019-12-01
Add initial Real Time manifests

yanirq:
The following manifests added to support RT features during deployment and boot:

[Node tuning operator](https://github.com/openshift/cluster-node-tuning-operator) manifests: tuned-network-latency, tuned-rt.
[Machine configuration](https://docs.openshift.com/container-platform/4.2/nodes/nodes/nodes-nodes-managing.html) to set kubeleteConfig parameters: kubeletconfig-worker-rt  and [setting](https://github.com/openshift/machine-config-operator) boot parameters:  mc-kargs-worker-rt.

> This will require node to be labled with `node-role.kubernetes.io/worker-rt=""` for these configurations could be applied


yanirq:
/cc @slintes @cynepco3hahue @marsik @fromanirh
This is an initial structure, since not much code is merged yet this is open for changes.
Note that setting rt kernel is not included yet and will probably align to the general deamonset machineconfig which is planned.

features/rt/manifests/01-tuned-rt.yaml
```diff
@@ -0,0 +1,40 @@
...
+      [net]
+      nf_conntrack_hashsize=131072
+
+      [sysctl]
+      kernel.hung_task_timeout_secs = 600
+      kernel.nmi_watchdog = 0
+      kernel.sched_rt_runtime_us = -1
+      vm.stat_interval = 10
+      kernel.timer_migration = 0
+      #vm.nr_hugepages=${hugepage_num}
```
cynepco3hahue: Can we remove the commented line?

features/rt/manifests/01-tuned-rt.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      #TODO - this should probably be removed since cpumask could be set only at an early stage (initramfs/initrd)
```
cynepco3hahue: Can we remove the commented line?

features/rt/manifests/05-mc-kargs-worker-rt.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+    labels:
+        machineconfiguration.openshift.io/role: worker-rt
+    name: 10-kargs-worker-rt
+spec:
+    kernelArguments:
+        - default_hugepagesz=1G
+        - hugepagesz=1G
+        - hugepages=32
+#        - isolcpus=2-15
```
cynepco3hahue: Can we remove the commented line?

jmencak: That's why this PR is called "initial", I guess.  isolcpus or similar way of pining to CPU cores will be needed and this PR doesn't address this.

cynepco3hahue: in this case, we can remove it and address it in the following PR
In general, I think we can provide a template that will be filled with relevant values because of such parameters as a number of huge pages or CPU that we want to isolate can vary from node to node.

fromanirh:
my take: add initial manifests first, clean up later (maybe in a separate PR) for easier review. Takes a little longer, but each step is very clear and focused.

slintes:
> Note that setting rt kernel is not included yet and will probably align to the general deamonset machineconfig

will be a machineconfig, yes. I am currently testing what I have, will add it in a follow up PR.

features/rt/manifests/05-kubeletconfig-worker-rt.yaml
```diff
@@ -0,0 +1,18 @@
...
+    matchLabels:
+      worker-rt: ""
+  kubeletConfig: 
+    systemReserved:
+      cpu: 1000m
+      memory: 500Mi
+    kubeReserved:
+      cpu: 1000m
+      memory: 500Mi
+    cpuManagerPolicy: static
```
MarSik: Do you think we should configure the default NUMA behavior right here too? At least topologyManagerPolicy: best-effort so basic NUMA alignment with no rejections are enabled?

yanirq: maybe we can, but i think more additions would be good to add in followup PRs

cynepco3hahue: Just be sure that you add the topology manager feature gate as well or we can just do it in the following PR.

features/rt/manifests/05-mc-kargs-worker-rt.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+    labels:
+        machineconfiguration.openshift.io/role: worker-rt
+    name: 10-kargs-worker-rt
+spec:
+    kernelArguments:
```
e-minguez: Just in case, I'd say the kernelArguments would be better if they are quoted (as in https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfiguration.md#kernelarguments )

yanirq: done

PR #21, opened 2019-12-02
comment disableselinux

ch-stark:
according to testing it should not be necessary to set selinux to permissive


simon3z:
@ch-stark maybe the PR title is misleading? Aren't you commenting disableselinux (vs uncommenting it). cc @sivak

simon3z:
cc @MarSik

PR #22, opened 2019-12-02
RT kernel

slintes:
~This is based on https://github.com/openshift-kni/baremetal-deploy/pull/20!~
edit: #20 moved to `performance` dir, I leave this in `rt`for now. We can merge later if needed

This adds the realtime kernel feature with
- a MachineConfig template
- a install script
- a generate script, which base64 encodes the install script and puts it into the MC template
- a Makefile, because it's easier to use than searching/calling the generate script (and whatever might follow)
- a initial README.md
- and the generated manifest

### DIFF
--- a/features/rt/hack/generate.sh
+++ b/features/rt/hack/generate.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -euo pipefail
+
+assetsDir=$(dirname "$0")/../assets
+manifestsDir=$(dirname "$0")/../manifests
+
+# Generate RT kernel MC
+echo "generating 06-mc-rtkernel-worker-rt.yaml"
+SCRIPT64="$(base64 -w 0 ${assetsDir}/rt-kernel-patch.sh)"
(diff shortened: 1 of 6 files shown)


simon3z:
@vladikr @cynepco3hahue please review. cc @MarSik
I think we should extract the kernel parameters to a single MachineConfig decoupled from rt.

slintes:
#20 moved to `performance` dir, I leave this in `rt` for now, so this can be reviewed independently. We can merge later if needed.

features/rt/hack/generate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
cynepco3hahue: hm can we move it to the root/hack directory I think everyone can use it?

slintes: I'd suggest to merge this and #26
and the see what we can consolidate
makes sense?
/cc @karmab

cynepco3hahue:
lgtm

slintes:
/cc @karmab 

anything missing here? :)

PR #38, opened 2019-12-05
Update install-steps.md

rlopez133:
complete revamp of the installation of IPI on barmetal by combining the great work of Mario Vazquez Cebrian David Vossel @schmaustech among others. The new source of truth (work in progress).

### DIFF
--- a/install-steps.md
+++ b/install-steps.md
@@ -1,193 +1,964 @@
-Ran the following commands on (worker-0 / provision host) on RHEL 8.1
+# Introduction
 
-Environment - TBD
+This write-up will guide you through the process of deploying a Baremetal IPI installation of Red Hat OpenShift 4.
+# Prerequisites
 
-Assumptions- TBD
+* 6 Physical servers (1 provision node, 3 master and 2 worker nodes)
+* Each server needs 2 NICs pre-configured. NIC1 for the private network and NIC2 for the external network. NIC interface names need to be identical. See [issue](https://github.com/openshift/installer/issues/2762)
+* Each server should have a RAID-1 configured and initialized
+* Each server must have IPMI configured
+* Each server must have DHCP setup for external NICs
+* Each server must have DNS setup for the API, wildcard applications
+* A DNS VIP is IP on the `baremetal` network is required for reservation. Reservation is done via our DHCP server (though not required).  
+* Optional - Include DNS entries for the external hostnames for each of the servers
+* Download a copy of your [Pull secret](https://cloud.redhat.com/openshift/install/metal/user-provisioned)
+* Append to the `pull-secret.txt` the [Pull secret](https://docs.google.com/document/d/1pWRtk7IbnfPo6cSDsopUMrxS22t3VJ2PuN39MJp9tHM/edit) with access to `registry.svc.ci.openshift.org` and `registry.redhat.io`
 
----
+## Networking Requirements
 
-```bash
-firewall-cmd --zone=public --add-service=http
-firewall-cmd --zone=public --permanent --add-service=http
+### NIC Configuration
 
+Each server requires two NICs for `provisioning` and `baremetal` network access 
+respectively. The `provisioning` network is a non-routable network used
+for provisioning the underlying operating system on each server that are part
+of the Red Hat OpenShift cluster. The `baremetal` network is a routable network
+used for external network access to the outside world. 
 
-subscription-manager register --username=<user> --password=<pass>
-subscription-manager repos --enable=rhel-8-for-x86_64-appstream-rpms --enable=rhel-8-for-x86_64-baseos-rpms
+NOTE: It is recommended that each NIC be on a separate VLAN. 
 
-#optional - if you gonna use preflight get ansible (to create install-config.yaml)
-subscription-manager repos --enable ansible-2-for-rhel-8-x86_64-rpms
-dnf install -y ansible
+### Network Ranges
 
-#optional - if you gonna use preflight, install jq
-dnf install -y jq
+The `provisioning` network is automatically assigned and configured using the
+`172.22.0.0/24` range. The `baremetal` (external) network assignment is to be provided
+by your network administrator. It is required that each server be assigned a 
+external IP address via a DHCP server. 
 
-dnf install -y libvirt qemu-kvm mkisofs
-systemctl start libvirtd
-systemctl enable libvirtd
+NOTE: Since the Red Hat OpenShift installation uses `ironic-dnsmasq` it is 
+important that no other DHCP server is running on the same broadcast domain.
 
-cat << EOF > default_storage.xml 
-<pool type='dir'>
-  <name>default</name>
-  <source>
-  </source>
-  <target>
-    <path>/var/lib/libvirt/images</path>
-    <permissions>
-      <mode>0711</mode>
-      <owner>0</owner>
-      <group>0</group>
-      <label>system_u:object_r:virt_image_t:s0</label>
-    </permissions>
-  </target>
-</pool>
-EOF
+### Reserved IPs on DHCP Server
 
-virsh pool-define default_storage.xml 
-virsh pool-start default
-virsh pool-autostart default
+Each server must have an IP reserved on the `baremetal` network. Additionally,
+3 additional IPs on the `baremetal` network are required for reservation. 
 
-nmcli connection add ifname provisioning type bridge con-name provisioning
-nmcli con add type bridge-slave ifname eno1 master provisioning
-nmcli connection add ifname baremetal type bridge con-name baremetal
-nmcli con add type bridge-slave ifname eno2 master baremetal
+* 1 IP for the API endpoint
+* 1 IP for the wildcard ingress endpoint
+* 1 IP for the name server
 
-#optional - verify you see the connections 
-nmcli con show
-
-#warning - if you are on an ssh terminal, you will kill connection running this
-nmcli con down eno2
-
-#warning - if you used dhclient to get an IP on eno2, kill the dhclient process
-ps -ef | grep dhclient
-kill -9 <id_of_dhclient_process>
-
-dhclient baremetal
-
-nmcli connection modify provisioning ipv4.addresses 172.22.0.1/24 ipv4.method manual
-nmcli con down provisioning
-nmcli con up provisioning 
-
-cat << EOF > provisioning.xml
-<network>
-  <name>provisioning</name>
-  <bridge name='provisioning'/>
-  <forward mode='bridge'/>
-</network>
-EOF
-
-virsh net-define provisioning.xml
-virsh net-start provisioning
-virsh net-autostart provisioning
-
-cat << EOF > baremetal.xml
-<network>
-  <name>baremetal</name>
-  <bridge name='baremetal'/>
-  <forward mode='bridge'/>
-</network>
-EOF
-
-virsh net-define baremetal.xml
-virsh net-start baremetal
-virsh net-autostart baremetal
-```
-
-     # optional - check the networks on virsh
-     virsh net-list
-     Name                 State      Autostart     Persistent
-     ----------------------------------------------------------
-     baremetal            active     yes           yes
-     default              active     yes           yes
-     provisioning         active     yes           yes
-
-```bash
-systemctl restart libvirtd
-
-# still need this? i cant remember why. 
-echo 'ZONE=libvirt' >> /etc/sysconfig/network-scripts/ifcfg-provisioning
-
-#Lets create a user (kni) so we aren't deploying as root and set ssh key
-echo "kni ALL=(root) NOPASSWD:ALL" | tee -a /etc/sudoers.d/kni
-chmod 0440 /etc/sudoers.d/kni
-su - kni -c "ssh-keygen -t rsa -f /home/kni/.ssh/id_rsa -N ''"
-
-#lets make sure python3 is default
-alternatives --set python /usr/bin/python3
+The following table shows a list of all items that require a reserved IP address
+on the `baremetal` network. Ensure a DHCP entry for each item listed on the
+table below is reserved and allocated at all times. Examples on using `dnsmasq`
+or a proper DHCP server to allocate these entries is shown later in this
+document. 
 
 
-#Get your pull-secret file on the system
+| Usage   |      Hostname      |  IP |
+|----------|-------------|------|
+| API | api.\<cluster-name\>.\<domain\> | \<ip\> |
+| Ingress LB (apps) |  *.apps.\<cluster-name\>.\<domain\>  | \<ip\> |
+| Nameserver | ns1.\<cluster-name\>.\<domain\> | \<ip\> |
+| Provisioning node | provisioner.\<cluster-name\>.<domain\> | \<ip\> |
+| Master-0 | \<cluster-name\>-master-0.<domain\> | \<ip\> |
+| Master-1 | \<cluster-name\>-master-1.<domain\> | \<ip\> |
+| Master-2 | \<cluster-name\>-master-2.<domain\> | \<ip\> |
+| Worker-0 | \<cluster-name\>-worker-0.<domain\> | \<ip\> |
+| Worker-1 | \<cluster-name\>-worker-1.<domain\> | \<ip\> |
+
+### DNS Server
+
+A subzone is required on the \<domain\> that is to be used. The subzone name
+is recommended to be the \<cluster-name\> as shown in the table above. Examples
+on using `dnsmasq` or a proper DNS server to allocate these DNS resolution entries
+is shown later in this document. 
+
+<!--
+
 TODO
+### Diagram 
 
-# Get the oc binary and place oc in /usr/local/bin
+![Network Diagram](diagrams/bos-server-net.png)
+-->
+# Installation Flow
 
-export VERSION=$(curl -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/latest/release.txt | grep 'Name:' | awk -F: '{print $2}' | xargs)
-export RELEASE_IMAGE=$(curl -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/latest/release.txt | grep 'Pull From: quay.io' | awk -F ' ' '{print $3}' | xargs)
-export cmd=openshift-baremetal-install
-export pullsecret_file=pull-secret.txt
-export extract_dir=$(pwd)
-curl -s https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/latest/openshift-client-linux-$VERSION.tar.gz | tar zxvf - oc
-sudo cp oc /usr/local/bin/
-```
+1. Installation of RHEL 8.1 on one of the 6 servers to be used as our provisioning node
+2. (Option 1) Configuration of DNS and DHCP entries
+3. (Option 2) Configuration of DNS and DHCP entries using `dnsmasq`
+4. Provisioning node to provision a 3 master, 2 worker Red Hat OpenShift cluster
+5. Once the cluster is up and running, re-provision the provisioning node as a worker node (worker-2) using `ironic`
+
+# Configuring Servers
+
+Each server requires the following configuration for proper installation. 
+
+WARNING: A mismatch between servers will cause an installation failure.
+
+While the servers that are used in your environment may contain additional
+NICs. For the purposes of installation, we are only focused on the following: 
+
+| NIC   | NETWORK | VLAN |
+|----------|-------------|------|
+| NIC1 | `provisioning`| \<provisioning-vlan\> |
+| NIC2 | `baremetal`| \<baremetal-vlan\> |
+
+While installation of RHEL 8.1 on the **provisioning** node will vary within
+environments, the reasoning
+for NIC2 to have PXE-enabled is to ensure easily installation using a local
+satellite server. As mentioned prior, NIC1 is a non-routable network (`provisioning`)
+that is only to be used for the installation of the Red Hat OpenShift cluster.
+
+| PXE | Boot Order |
+|----------|-------|
+| NIC1 PXE-enabled (`provisioning` network) | 1 |
+| NIC2 PXE-enabled (`baremetal` network) | 2 |
+
+NOTE: PXE has been disabled on all other NICs.
+
+The **master** and **worker** nodes have been configured as follows:
+
+| PXE | Boot Order |
+|----------|-------|
+| NIC1 PXE-enabled (`provisioning` network) | 1 |
+
+NOTE: PXE has been disabled on all other NICs.
+
+## Out of Band Management
+
+Each server is required to have access to out of band management. The
+provisioning node will require access to the out of band management network for
+a successful OpenShift 4 installation. The out of band management setup is
+out of scope for this document, however, it is recommended that a separate 
+management network be created for best practices. This is not applicable, using
+either the `provisioning` network or `baremetal` network are other options as
+well. 
+
+## Required Data for Installation
+
+Prior to the installation of the Red Hat OpenShift cluster, it is required to
+gather the following information from **all** servers. The list includes:
+
+* Out-of-Band Management IP
+    * Examples
+        * Dell (iDRAC) IP
+        * HP (iLO) IP
+* NIC1 (`provisioning`) MAC address
+* NIC2 (`baremetal`) MAC address
+
+# Reserve IPs for the VIPs and Nodes
+
+Previously, it was mentioned we would need the following IP address reserved
+on the `baremetal` network. The list included:
+
+* IP for the api endpoint
+* IP for the wildcard ingress endpoint
+* IP for the nameserver
+* IP for each node 
+
+The total amount of IPs required for reservation is 9 - 6 IPs for the nodes
+and 3 IPs for the API, wildcard ingress endpoint and the nameserver. It is
+important to contact your network administrator to reserve these nine IPs to 
+ensure no conflict on the network. 
+
+# Create DNS records on a DNS server (Option 1)
+
+Option 1 should be used if access to the appropriate DNS server for the `baremetal`
+network is accessible or a request to your network admin to create the DNS records
+is an option. If not an option, skip this section and move to section Create
+DNS records using `dnsmasq` (Option 2).
+
+First, create a subzone with the name of the cluster that is going to be used on
+your domain. For clarity in the example, the domain used is `example.com` and the
+cluster name used is `openshift`. Ensure to change these according to your 
+environment specifics. 
+
+1. Login to the DNS server via `ssh`
+2. Suspend updates to all dynamic zones: `rndc freeze`
+3. Edit /var/named/dynamic/example.com
+
+    ~~~
+    $ORIGIN openshift.example.com.
+    $TTL 300        ; 5 minutes
+    @  IN  SOA  dns1.example.com.  hostmaster.example.com. (
+           2001062501  ; serial
+           21600       ; refresh after 6 hours
+           3600        ; retry after 1 hour
+           604800      ; expire after 1 week
+           86400 )     ; minimum TTL of 1 day
+    ;
+    api                     A       <api-ip>
+    ns1                     A       <dns-vip-ip>
+    $ORIGIN apps.openshift.example.com.
+    *                       A       <wildcard-ingress-lb-ip>
+    $ORIGIN openshift.example.com.
+    provisioner             A       <NIC2-ip-of-provision>
+    openshift-master-0      A       <NIC2-ip-of-master-0>
+    openshift-master-1      A       <NIC2-ip-of-master-1>
+    openshift-master-2      A       <NIC2-ip-of-master-2>
+    openshift-worker-0      A       <NIC2-ip-of-worker-0>
+    openshift-worker-1      A       <NIC2-ip-of-worker-1>
+    ~~~
+4. Increase the SERIAL value by 1
+5. Edit /var/named/dynamic/1.0.10.in-addr.arpa
+
+    ~~~
+    $ORIGIN 1.0.10.in-addr.arpa.
+    $TTL 300
+    @  IN  SOA  dns1.example.com.  hostmaster.example.com. (
+           2001062501  ; serial
+           21600       ; refresh after 6 hours
+           3600        ; retry after 1 hour
+           604800      ; expire after 1 week
+           86400 )     ; minimum TTL of 1 day
+    ;
+    126 IN      PTR      provisioner.openshift.example.com.
+    127	IN 	PTR	openshift-master-0.openshift.example.com.
+    128	IN 	PTR	openshift-master-1.openshift.example.com.
+    129	IN 	PTR	openshift-master-2.openshift.example.com.
+    130	IN 	PTR	openshift-worker-0.openshift.example.com.
+    131	IN 	PTR	openshift-worker-1.openshift.example.com.
+    132 IN      PTR     api.openshift.example.com.
+    133 IN      PTR     ns1.openshift.example.com.
+    ~~~
+    
+    NOTE: In this example IP address `10.0.1.126-133` are pointed to the
+    correspoding fully qualified domain name. 
+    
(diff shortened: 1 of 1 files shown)


install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+
+1. Create the sample ConfigMap `metal3-config.yaml.sample`
+   ~~~yaml
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: metal3-config
+      namespace: openshift-machine-api
+    data:
+      cache_url: ${RHCOS_URI}
```
e-minguez: I think this is better to set it to `''` to jump directly to download the rhcos file

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+      username: ca2vdAo=
+      password: MWAwTWdtdC0K
+    ---
+    apiVersion: metal3.io/v1alpha1
+    kind: BareMetalHost
+    metadata:
+      name: openshift-worker-2
+    spec:
+      online: true
+      bootMACAddress: <NIC1-mac-address>
```
e-minguez: I've always needed to set `hardwareProfile: unknown` in the workers to make this work

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+            bootMACAddress: <NIC1-mac-address
+            hardwareProfile: default
+          - name: openshift-worker-1
+            role: worker
+            bmc:
+              address: ipmi://<out-of-band-ip>
+              username: <user>
+              password: <password>
+            bootMACAddress: <NIC1-mac-address
+            hardwareProfile: default
```
e-minguez: I've always needed to set `hardwareProfile: unknown` in the workers to make this work

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+    NAME                 STATUS   PROVISIONING STATUS   CONSUMER   BMC                 HARDWARE PROFILE   ONLINE   ERROR
+    openshift-worker-2   OK       inspecting                       ipmi://<out-of-band-ip>                      true     
+    ~~~
+7. Once the inspection finishes the node will be ready to be provisioned
+
+    ~~~sh
+    oc -n openshift-machine-api get bmh openshift-worker-2
+
+    NAME                 STATUS   PROVISIONING STATUS   CONSUMER   BMC                 HARDWARE PROFILE   ONLINE   ERROR
+    openshift-worker-2   OK       ready                            ipmi://<out-of-band-ip>   unknown            true
```
e-minguez: I'd say to add instructions on how to see the ironic stuff under the hood:

* Create a clouds.yaml file with the following content:
```
clouds:
  metal3-bootstrap:
    auth_type: none
    baremetal_endpoint_override: http://172.22.0.2:6385
    baremetal_introspection_endpoint_override: http://172.22.0.2:5050
  metal3:
    auth_type: none
    baremetal_endpoint_override: http://172.22.0.3:6385
    baremetal_introspection_endpoint_override: http://172.22.0.3:5050
```

* Install the openstack and ironic cli (I prefer to set a virtualenv and pip on that venv)
* Then query ironic as:
```
export OS_CLOUD="metal3"
openstack baremetal node list
```

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+   ~~~
+
+4. Run the OpenShift Installer
+    ~~~sh
+    ./openshift-baremetal-install --dir ~/clusterconfigs --log-level debug create cluster
+    ~~~
+
+## Fix Metal3 ConfigMap
+
+With the installation complete. The focus shifts to fixing the metal3 ConfigMap
```
e-minguez: The installation won't finish until the metal3-operator is up and running, so it is required to create the ConfigMap during the installation:
```
export KUBECONFIG=~/clusterconfigs/auth/kubeconfig
oc create -f metal3-configmap.yaml
```

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+
+3. Ensure all baremetal nodes are powered off prior to installing the OpenShift cluster
+   ~~~sh
+   ipmitool -I lanplus -U <user> -P <password> -H <management-server-ip> -p 6230 power off
+   ~~~
+
+4. Run the OpenShift Installer
+    ~~~sh
+    ./openshift-baremetal-install --dir ~/clusterconfigs --log-level debug create cluster
+    ~~~
```
e-minguez: If deploying masters + workers it will eventually timeout... in that case, the `wait-for install-complete` will be required if you want to see a successful finish.

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+2. Start `firewalld`, enable the `http` service, enable port 5000.
+   ~~~sh
+   systemctl start firewalld
+   firewall-cmd --zone=public --add-service=http --permanent
+   firewall-cmd --zone=public --permanent --add-service=http --permanent
+   firewall-cmd --add-port=5000/tcp --zone=libvirt  --permanent
+   firewall-cmd --add-port=5000/tcp --zone=public   --permanent
+   firewall-cmd --reload
+   ~~~
+3. Disable SELinux
```
karmab: i think disabling selinux is not needed anymore?

install-steps.md
mvazquezc: Just for reference: I've tried without setting up the hardwareProfile, and that will end up in hardwareProfile being set to unknown, that being said, configuring it to unknow by default can be a good idea.

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+    setenforce 0
+    ~~~
+4. Register your environment using `subscription-manager`
+   ~~~sh
+   subscription-manager register --username=<user> --password=<pass>
+   subscription-manager repos --enable=rhel-8-for-x86_64-appstream-rpms --enable=rhel-8-for-x86_64-baseos-rpms
+   ~~~
+4. Install the following packages 
+   ~~~sh
+   dnf install -y libvirt qemu-kvm mkisofs python3-devel python3-pip gcc ipmitool
```
karmab: pip and gcc are not needed

install-steps.md
mvazquezc: What about running openstack client from a container with podman?

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+3. Ensure all baremetal nodes are powered off prior to installing the OpenShift cluster
+   ~~~sh
+   ipmitool -I lanplus -U <user> -P <password> -H <management-server-ip> -p 6230 power off
+   ~~~
+
+4. Run the OpenShift Installer
+    ~~~sh
+    ./openshift-baremetal-install --dir ~/clusterconfigs --log-level debug create cluster
+    ~~~
+
```
karmab: we need to have openstack client on the machine running the install to help troubleshooting and preconfigured with clouds.yaml so it can be used during bootstrap and afterwards

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+      rhcos_image_url: ${RHCOS_PATH}${RHCOS_URI}
+    ~~~
+    NOTE: The `provision_ip` should be modified to an available IP on the `provision` network. The default is `172.22.0.3`
+
+2. Create the final ConfigMap
+    ~~~sh
+    export COMMIT_ID=$(./openshift-baremetal-install version | grep '^built from commit' | awk '{print $4}')
+    export RHCOS_URI=$(curl -s -S https://raw.githubusercontent.com/openshift/installer/$COMMIT_ID/data/data/rhcos.json | jq .images.openstack.path | sed 's/"//g')
+    export RHCOS_PATH=$(curl -s -S https://raw.githubusercontent.com/openshift/installer/$COMMIT_ID/data/data/rhcos.json | jq .baseURI | sed 's/"//g')
+    envsubst < metal3-config.yaml.sample > metal3-config.yaml
```
karmab: i would inject this configmap prior to the deployment, so no fix is required after launching the openshift.install command

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+the node, it is recommended to backup some existing files.
+
+~~~sh
+# Tar the cluster configs folder and download it to your laptop
+tar cvfz clusterconfig.tar.gz ~/clusterconfig
+# Copy the Private part for the SSH Key configured on the install-config.yaml to your laptop
+tar cvfz clusterconfigsh.tar.gz ~/.ssh/id_rsa*
+~~~
+
+# Deploy Worker Node
```
karmab: i would deploy the worker during the initial installation, by setting them properly in the install-config.yaml

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
-alternatives --set python /usr/bin/python3
-
-
-#Get your pull-secret file on the system
+# Introduction
+
+This write-up will guide you through the process of deploying a Baremetal IPI installation of Red Hat OpenShift 4.
+# Prerequisites
+
+* 6 Physical servers (1 provision node, 3 master and 2 worker nodes)
```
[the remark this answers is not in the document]
e-minguez: Or patch the ingresscontroller to set just a single replica:

```
oc patch -n openshift-ingress-operator ingresscontroller/default --patch '{"spec":{"replicas": 1}}' --type=merge
```

install-steps.md
DirectedSoul1: I have made similar observances.

install-steps.md
e-minguez: > What about running openstack client from a container with podman?

It is a bit tricky because AFAIK the openstack-client rpm is only available in the OSP channels so I'm not sure if we provide an openstack-client container. That being said, it is similar in terms of (in)security to install the tool via pip than pulling an insecure container image :D

rlopez133: I know Steve Hardy had something going here regarding osp client in a container: https://github.com/openshift/ironic-image/pull/17

install-steps.md
rlopez133: I initially thought this as well, I've seen conflicting messages. We do need to confirm 1 way or the other.

install-steps.md
rlopez133: OK - will make that change.

install-steps.md
rlopez133: What is the best way to handle this? Do we want to add the OSP repos so we can officially install the client? Do we want to do this with a container? See: https://github.com/openshift/ironic-image/pull/17 , Do we want to do it with pip? I see this as the least favorable due to support.

install-steps.md
rlopez133: @karmab ++

install-steps.md
rlopez133: I actually need to change the title. This doc already deploys workers via install-config.yaml . This is to convert the provision node into a worker node after initial deployment.

e-minguez:
I've added some comments, but also I think the document is quite long. I'd say to maybe split it in different docs such as prereq.md, add-worker.md, etc. and a general install.md with just the install steps.
My 2 cents.

rlopez133:
@e-minguez - im thinking a better approach will be to have this long version and then have a "quick start" of sorts that is condensed aggressively. If someone doesn't understand a particular section, they can then just reference the long version. My 2 cents.

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+    nmcli con delete "$PROV_CONN"
+    nmcli con add type bridge ifname provisioning autoconnect yes con-name provisioning stp off
+    nmcli con modify provisioning ipv4.addresses 172.22.0.1/24 ipv4.method manual
+    nmcli con add type bridge-slave autoconnect yes con-name "$PROV_CONN" ifname "$PROV_CONN" master provisioning
+    nmcli con delete "$PUB_CONN"
+    nmcli con add type bridge ifname baremetal autoconnect yes con-name baremetal stp off
+    nmcli con add type bridge-slave autoconnect yes con-name "$PUB_CONN" ifname "$PUB_CONN" master baremetal
+    systemctl restart NetworkManager
+    systemctl restart libvirtd
+    EOS
```
schmaustech: So I do it this way:
    nmcli connection add ifname provisioning type bridge con-name provisioning
    nmcli con add type bridge-slave ifname "$PROV_CONN" master provisioning
    nmcli connection add ifname baremetal type bridge con-name baremetal
    nmcli con add type bridge-slave ifname "$MAIN_CONN" master baremetal
    nmcli con down "$MAIN_CONN";pkill dhclient;dhclient baremetal
    nmcli connection modify provisioning ipv4.addresses 172.22.0.1/24 ipv4.method manual
    nmcli con down provisioning
    nmcli con up provisioning
No embedded script - and by passing the pkill dclient - dhclient line all at once the connection is not lost - there will be a slight delay and then it re-establishes automatically

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+5. Start and enable the `libvirtd` service
+   ~~~sh
+   systemctl start libvirtd
+   systemctl enable libvirtd --now
+   ~~~
+6. Create the `default` storage pool and start it.
+   ~~~sh
+   virsh pool-define-as --name default --type dir --target /var/lib/libvirt/images
+   virsh pool-start default
+   virsh pool-autostart default
```
schmaustech: Are we adding the user that will be doing the deploy to the libvirt group?
usermod --append --groups libvirt <cluster deploy user>
Otherwise when the bootstrap VM is attempted you get a polkit auth error.

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
-
-
-#Get your pull-secret file on the system
+# Introduction
+
+This write-up will guide you through the process of deploying a Baremetal IPI installation of Red Hat OpenShift 4.
+# Prerequisites
+
+* 6 Physical servers (1 provision node, 3 master and 2 worker nodes)
+* Each server needs 2 NICs pre-configured. NIC1 for the private network and NIC2 for the external network.
```
dhellmann: It's also necessary for the 3 master hosts to have similar enough hardware that the interface names for the NICs are identical. See https://github.com/openshift/installer/issues/2762

install-steps.md
rlopez133: I'll make sure to add this

install-steps.md
rlopez133: I'll make sure to make a note of that in prereqs.

install-steps.md
rlopez133: confirmed it. will remove disable selinux section

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+
+The following steps need to be performed in order to prepare the environment. 
+
+1. Login into the provision node via `ssh`
+2. Start `firewalld`, enable the `http` service, enable port 5000.
+   ~~~sh
+   systemctl start firewalld
+   firewall-cmd --zone=public --add-service=http --permanent
+   firewall-cmd --zone=public --permanent --add-service=http --permanent
+   firewall-cmd --add-port=5000/tcp --zone=libvirt  --permanent
```
rlopez133: This will not work because the libvirt package is yet to be installed here. This must go after libvirt is installed.

rlopez133: This will not work because the libvirt package is yet to be installed here. This must go after libvirt is installed.

install-steps.md
rlopez133: confirmed dont need python3-pip or gcc

install-steps.md
rlopez133: confirmed requires
```
usermod --append --groups libvirt <user>
```

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+    ~~~
+8. `ssh` back into your terminal session
+9. Verify the connection bridges have been properly created
+    ~~~sh
+    nmcli con show
+    ~~~
+10. Create a user (i.e `kni`) to deploy as non-root
+    ~~~sh
+    echo "kni ALL=(root) NOPASSWD:ALL" | tee -a /etc/sudoers.d/kni
+    chmod 0440 /etc/sudoers.d/kni
```
rlopez133: This needs to be one of the first steps done and then make all the other commands use `sudo`

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+11. Create an `ssh` key for the new user (i.e. kni)
+    ~~~sh
+    su - kni -c "ssh-keygen -t rsa -f /home/kni/.ssh/id_rsa -N ''"
+    ~~~
+12. Ensure that `python3` is default python
+    ~~~sh
+    alternatives --set python /usr/bin/python3
```
rlopez133: not required as if you verify `python --version` it shows version 3.x

install-steps.md
rlopez133: require the package `jq` to be installed due to capturing the RHCOS_URI via command later in the document.

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+
+2. Create a directory to store cluster configs
+    ~~~sh
+    mkdir ~/clusterconfigs
+    cp install-config.yaml ~/clusterconfigs
+    ~~~
+
+3. Ensure all baremetal nodes are powered off prior to installing the OpenShift cluster
+   ~~~sh
+   ipmitool -I lanplus -U <user> -P <password> -H <management-server-ip> -p 6230 power off
```
rlopez133: When I do this command I dont require port 6230. @schmaustech @e-minguez @karmab - I believe this isn't needed. Command works without.

install-steps.md
rlopez133: @e-minguez I was able to get a successful deployment without the need for `wait-for install-complete` Spoke to @schmaustech and he has also gotten installations to work without the need of this.

install-steps.md
rlopez133: confirmed -- I will also add details that it requires the API to be up and running.

install-steps.md
rlopez133: @davidvossel @karmab - today I went through this and was also trying to figure out how would you add this fix prior to launching the opneshift install command?

install-steps.md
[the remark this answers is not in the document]
rlopez133: > we can work around the 2 workers nodes requirement by providing a custom IngressController where we force the role where to put the routers and their replica number
> 
> ```
> apiVersion: operator.openshift.io/v1
> kind: IngressController
> metadata:
>   name: default
>   namespace: openshift-ingress-operator
> spec:
>   replicas: {{ replicas }}
>   endpointPublishingStrategy:
>     type: HostNetwork
>   nodePlacement:
>     nodeSelector:
>       matchLabels:
>         node-role.kubernetes.io/{{ role }}: ""
> ```

I don't think the goal isto work around the issue. The goal here is to provide what the 'best practice' should be.

install-steps.md
rlopez133: fixed.

install-steps.md
rlopez133: added `hardwareProfile: unknown` to the workers

install-steps.md
rlopez133: Going to resolve this as it was not needed.

install-steps.md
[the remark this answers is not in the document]
rlopez133: @sreichar thoughts on this?

install-steps.md
rlopez133: @karmab @schmaustech @sreichar - thoughts here?

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+
+* 6 Physical servers (1 provision node, 3 master and 2 worker nodes)
+* Each server needs 2 NICs pre-configured. NIC1 for the private network and NIC2 for the external network. 
+* Each server should have a RAID-1 configured and initialized
+* Each server must have IPMI configured
+* Each server must have DHCP setup for external NICs
+* Each server must have DNS setup for the API, wildcard applications and nameserver (ns1) 
+* Optional - Include DNS entries for the external hostnames for each of the servers
+* Download a copy of your [Pull secret](https://cloud.redhat.com/openshift/install/metal/user-provisioned)
+* Append to the `pull-secret.txt` the [Pull secret](https://docs.google.com/document/d/1pWRtk7IbnfPo6cSDsopUMrxS22t3VJ2PuN39MJp9tHM/edit) with access to `registry.svc.ci.openshift.org` and `registry.redhat.io`
```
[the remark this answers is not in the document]
rlopez133: > I'm not sure this document is open for people outside the organization.

What are you trying to imply here? That the google doc doesn't make sense to be part of this?

install-steps.md
rlopez133: I had issues with the script as is. I will investigate your way further.

install-steps.md
e-minguez: I think it would be nice to included as optional or a just in case... I've hit the timeout all the times I've deployed in baremetal :man_shrugging:

install-steps.md
rlopez133: since no comment and I was able to confirm it worked without I'm resolving.

install-steps.md
schmaustech: I do the following:
        OPENSHIFT_INSTALLER=/usr/local/bin/openshift-install
        OPENSHIFT_INSTALL_COMMIT=$($OPENSHIFT_INSTALLER version | grep commit | cut -d' ' -f4)
        OPENSHIFT_INSTALLER_RHCOS=${OPENSHIFT_INSTALLER_RHCOS:-https://raw.githubusercontent.com/openshift/installer/$OPENSHIFT_INSTALL_COMMIT/data/data/rhcos.json}
        RHCOS_IMAGE_JSON=$(curl "${OPENSHIFT_INSTALLER_RHCOS}")
        export RHCOS_INSTALLER_IMAGE_URL=$(echo "${RHCOS_IMAGE_JSON}" | jq -r '.baseURI + .images.openstack.path')
        export RHCOS_IMAGE_URL=${RHCOS_IMAGE_URL:-${RHCOS_INSTALLER_IMAGE_URL}}
        sed -i "s#RHCOS_IMAGE_URL#${RHCOS_IMAGE_URL}#" /home/bschmaus/metal3-config.yaml
/usr/local/bin/openshift-install --dir=ocp create manifests
cp /home/bschmaus/metal3-config.yaml ocp/openshift/99_metal3-config.yaml
/usr/local/bin/openshift-install --dir=ocp --log-level debug create cluster

install-steps.md
schmaustech: pip is normally not supported from a real product perspective (think GA here).  I would think we would want to obtain the openstack client from a proper package channel

install-steps.md
```diff
@@ -1,193 +1,821 @@
...
+
+This write-up will guide you through the process of deploying a Baremetal IPI installation of Red Hat OpenShift 4.
+# Prerequisites
+
+* 6 Physical servers (1 provision node, 3 master and 2 worker nodes)
+* Each server needs 2 NICs pre-configured. NIC1 for the private network and NIC2 for the external network. 
+* Each server should have a RAID-1 configured and initialized
+* Each server must have IPMI configured
+* Each server must have DHCP setup for external NICs
+* Each server must have DNS setup for the API, wildcard applications and nameserver (ns1)
```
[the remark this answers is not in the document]
schmaustech: ns1 is not needed but DNS_VIP IP is.

install-steps.md
```diff
@@ -1,193 +1,850 @@
...
+table below is reserved and allocated at all times. Examples on using `dnsmasq`
+or a proper DHCP server to allocate these entries is shown later in this
+document. 
+
+
+| Usage   |      Hostname      |  IP |
+|----------|-------------|------|
+| API | api.\<cluster-name\>.\<domain\> | \<ip\> |
+| Ingress LB (apps) |  *.apps.\<cluster-name\>.\<domain\>  | \<ip\> |
+| Nameserver | ns1.\<cluster-name\>.<domain\> | \<ip\> |
```
rlopez133: can be removed, just need to make sure i have a reserved IP

install-steps.md
```diff
@@ -1,193 +1,850 @@
...
+    $TTL 300        ; 5 minutes
+    @  IN  SOA  dns1.example.com.  hostmaster.example.com. (
+           2001062501  ; serial
+           21600       ; refresh after 6 hours
+           3600        ; retry after 1 hour
+           604800      ; expire after 1 week
+           86400 )     ; minimum TTL of 1 day
+    ;
+    api                     A       <api-ip>
+    ns1                     A       <namserver-ip>
```
rlopez133: remove

install-steps.md
```diff
@@ -1,193 +1,850 @@
...
+media. 
+
+# Preparing the Provision node for OpenShift Install
+
+The following steps need to be performed in order to prepare the environment. 
+
+1. Login into the provision node via `ssh`
+2. Create a user (i.e `kni`) to deploy as non-root and provide that user `sudo` privileges
+    ~~~sh
+    useradd `kni`
```
rlopez133: remove ticks

install-steps.md
```diff
@@ -1,193 +1,850 @@
...
+    echo "kni ALL=(root) NOPASSWD:ALL" | tee -a /etc/sudoers.d/kni
+    chmod 0440 /etc/sudoers.d/kni
+3. Modify the user to add the `libvirt` group to the newly created user (i.e. `kni`)
+   ~~~sh
+   usermod --append --groups libvirt <user>
```
rlopez133: this needs to be run after libvirt has been installed

install-steps.md
```diff
@@ -1,193 +1,850 @@
...
+   ~~~
+6. Install the following packages 
+   ~~~sh
+   dnf install -y libvirt qemu-kvm mkisofs python3-devel jq ipmitool
+   ~~~
+7. Start `firewalld`, enable the `http` service, enable port 5000.
+   ~~~sh
+   systemctl start firewalld
+   firewall-cmd --zone=public --add-service=http --permanent
+   firewall-cmd --zone=public --permanent --add-service=http --permanent
```
rlopez133: remove

install-steps.md
```diff
@@ -1,193 +1,850 @@
...
+            bootMACAddress: <NIC1-mac-address>
+            hardwareProfile: default
+          - name: openshift-master-2
+            role: master
+            bmc:
+              address: ipmi://<out-of-band-ip>
+              username: <user>
+              password: <password>
+            bootMACAddress: <NIC1-mac-address
+            hardwareProfile: unknown
```
dhellmann: The hardwareProfile settings don't match throughout this example. I'm not sure if the installer uses different values ("default") from metal3 ("unknown"), so I don't know which to suggest you use. You could probably leave them out entirely except in cases where you need to be explicit to ensure the right disk is selected for the deployment.

rlopez133: @dh

> The hardwareProfile settings don't match throughout this example. I'm not sure if the installer uses different values ("default") from metal3 ("unknown"), so I don't know which to suggest you use. You could probably leave them out entirely except in cases where you need to be explicit to ensure the right disk is selected for the deployment.

@dhellmann my understanding is the master nodes require default and the worker nodes require unknown from what @e-minguez told me. I also tested an env this way where masters were default and workers were unknown and it worked for me.

dhellmann: OK. Master 0 and 1 are using "default" and master 2 is using "unknown". Worker 0 is using "unknown" and worker 1 is using "default".

install-steps.md
rlopez133: @schmaustech I added this in the doc. Ty for the steps.

install-steps.md
rlopez133: nice catch. fixed.

install-steps.md
rlopez133: Does this piece really belong in the installation doc? Should this be a different doc regarding troubleshooting? @e-minguez @karmab

rlopez133:
@e-minguez @karmab @davidvossel @dhellmann @sreichar 
I've made all the edits required. I believe the document is ready for merge. 

What is outstanding right now but I don't believe it should stop us from merging. 

1. @karmab  mentioned "we can work around the 2 workers nodes requirement by providing a custom IngressController where we force the role where to put the routers and their replica number" 
My answer: I don't think the goal is to work around the issue. The goal here is to provide what the 'best practice' should be.

2. @e-minguez mentioned "I'd say to add instructions on how to see the ironic stuff under the hood: Create a clouds.yaml file with the following content:..."
My answer: I feel these instructions would fall under some sort of troubleshooting section as being able to use the openstack client commands really only matter if there is an issue. ( @karmab had a similar comment regarding the osp client)

3. @davidvossel mentioned "One thing that I've been thinking about is a way to represent all of the input data associated with preparing [the environment ].... " 
My answer: While creating scripts to simplify these manual steps is something we should do, I don't think this changes what we are trying to accomplish in this PR which is to merge a standard doc on how to deploy. 

With that, @davidvossel would you be the one merging this in or someone else?

e-minguez:
> ...
>     2. @e-minguez mentioned "I'd say to add instructions on how to see the ironic stuff under the hood: Create a clouds.yaml file with the following content:..."
>        My answer: I feel these instructions would fall under some sort of troubleshooting section as being able to use the openstack client commands really only matter if there is an issue. ( @karmab had a similar comment regarding the osp client)
> ...

Ok to me. I think a troubleshooting doc would be really needed.

install-steps.md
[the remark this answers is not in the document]
karmab: i think the best practice should be a deployed environment

install-steps.md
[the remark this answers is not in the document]
karmab: i think the ci secret is not needed

install-steps.md
karmab: i think the idea from dev scripts folks is to bundle it in a container and use that

install-steps.md
karmab: it can go in a troubleshooting doc indeed

dhellmann:
> 2. @e-minguez mentioned "I'd say to add instructions on how to see the ironic stuff under the hood: Create a clouds.yaml file with the following content:..."
>    My answer: I feel these instructions would fall under some sort of troubleshooting section as being able to use the openstack client commands really only matter if there is an issue. ( @karmab had a similar comment regarding the osp client)

I agree, and want to go further and remind everyone that Ironic is an implementation detail. If we have to look at Ironic directly to understand what the system is doing then that is a bug in metal3's API, logging, or metric collection. A user should be able to debug using normal OpenShift tools, without installing the OpenStack CLI. I know we're not there yet, but let's not emphasize debugging procedures that we don't want to be using long term.

install-steps.md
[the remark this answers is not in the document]
rlopez133: > i think the best practice should be a deployed environment

Not sure what you mean by this comment. Can you elaborate? The doc as it stands does deploy an environment.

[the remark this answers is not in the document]
karmab: my point is that the instructions should work for an arbitrary number of workers, since
- most people wiling to use those instructions won't have that much hardware available.
- potential end users might have a need for a single worker or none.

Granted that, we should detail the steps needed to overcome this current ingresscontroller limitation (ever using edu's trick or mine).

install-steps.md
```diff
@@ -1,193 +1,964 @@
...
+   INFO Consuming Install Config from target directory 
+   WARNING Making control-plane schedulable by setting MastersSchedulable to true for Scheduler cluster settings 
+   WARNING Discarding the Openshift Manifests that was provided in the target directory because its dependencies are dirty and it needs to be regenerated 
+   ~~~
+7. Copy the `metal3-config.yaml` to the `clusterconfigs/openshift` directory
+   ~~~sh
+   cp ~/metal3-config.yaml clusterconfigs/openshift/99_metal3-config.yaml
+   ~~~
+
+## Deploying Routers on Worker Nodes
```
rlopez133: @karmab @e-minguez - ive updated the doc to reflect the deploying of routers

install-steps.md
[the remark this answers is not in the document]
rlopez133: > my point is that the instructions should work for an arbitrary number of workers, since
> 
> * most people wiling to use those instructions won't have that much hardware available.
> * potential end users might have a need for a single worker or none.
> 
> Granted that, we should detail the steps needed to overcome this current ingresscontroller limitation (ever using edu's trick or mine).

I've updated accordingly based on your feedback

install-steps.md
rlopez133: > i think the idea from dev scripts folks is to bundle it in a container and use that

This is gonna likely go into a troubleshoot doc not inside installer.

install-steps.md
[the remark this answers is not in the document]
rlopez133: > i think the ci secret is not needed

Have you tested to confirm?

PR #42, opened 2019-12-06
tests: bootstrap unit tests

fromanirh:
# Description
This patch adds the initial unit tests and their basic infrastructure.
Unit tests are mean to do simple, static checks, and to run (very)
often. Unit tests will check that configuration is syntactacally correct
(e.g. kernel args look sane, config file are parsable, base64-encoded
data can be decoded in something meaningful, any crossreference is
correct).

More elaborate testing (e.g. kernel tunables are actually applied,
isolated CPUs are really as such) will be done in e2e tests.

The tests are implemented in python, using pytest and pipenv, to make them
easy to write and to consume.

We use pipenv to easily have self-contantained, reproducible environment
on which we can run the python code.
We use pytest to have a complete, low-friction testing framework.

Each test should be put as close as possible to thing being tested,
using the `test_XXX.py` naming convention.

Any common utility shared among testing code should be put in the
`testlib.py` module. Code here should be added sparingly, and only
after an actual duplication is detected, to keep the module manageable.

A Makefile is provided to make it easy to run tests and to simplify
integration with the CI.

Usage:
1. run `make init` just once to prepare the python environment
   you can run `make init` repeatedly, but this has no further benefit
   once the initial setup is done
2. run `make test` any time you want to run the testsuite

References:
- pipenv docs: https://pipenv.kennethreitz.org
- CI integration tips: https://pipenv.kennethreitz.org/en/latest/advanced/#testing-projects
- pytest framework: https://docs.pytest.org/en/latest/

Fixes # (issue)
N/A

## Type of change

## Testing

The feature is self-testing

**Test Configuration**:

- Versions: master
- Hardware: not relevant

## Checklist

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+PYTHONPATH=.
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,6 @@
+init:
+	{ pipenv --version; } || pip install pipenv
+	pipenv install --dev
+
+test:
+	pipenv run pytest
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,13 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+pyyaml = "*"
+pytest = "*"
+
+[requires]
+python_version = "3.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,117 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "f0e48359fb4966f6753eed05fd6c01283bf47f3fe3d7e66258cb696022e692a0"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.7"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "attrs": {
+            "hashes": [
+                "sha256:08a96c641c3a74e44eb59afb61a24f2cb9f4d7188748e76ba4bb5edfa3cb7d1c",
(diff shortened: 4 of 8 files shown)


fromanirh:
> /cc @fedepaol
> 
> we need to align on a common test suit. i'm seeing ginko, pytest, and bash show up.
> 
> In order to further align with openshift, I'd suggest standardizing on ginko/gomega. That's what origin uses for the conformance tests. https://github.com/openshift/origin/tree/master/test/extended

I don't mind using ginkgo/gomega. My thought process was explained in the commit message, but if we settle for golang I'd happily rewrite this PR.
BTW, I think there is room for more than one language (e.g. python + golang, golang + shell) in our test toolbox.

simon3z:
> In order to further align with openshift, I'd suggest standardizing on ginko/gomega. That's what origin uses for the conformance tests. https://github.com/openshift/origin/tree/master/test/extended

+1

fromanirh:
Fine, makes sense. Since we need to rewrite them, do we want to keep these simple tests as start, or should we aim for something like functests/e2e tests?

MarSik:
@davidvossel  @simon3z About golang based tests... I must say I strongly prefer Python for anything that does test parsing as @fromanirh did and wanted to do here. The text files are pretty much free form and code written in dynamic language is just so much more readable when handling those.

So before you force everybody to a single platform, please sit down and think about the purpose and readability of the code that will be written.

Think about such simple lines like: `manifest_tree['spec']['kubeletConfig']` - in golang this would require vendoring of all the structures in the right versions (might be ok or might be an upgrade hassle like with webhooks) or using interface[string]map and the type casting syntax..

If the answer is still golang, then sure why now. Just do not blindly appeal to authority here.

fromanirh:
golang counterpart here: https://github.com/openshift-kni/baremetal-deploy/pull/52

fromanirh:
it seems we prefer to go in the direction taken inhttps://github.com/openshift-kni/baremetal-deploy/pull/52

PR #43, opened 2019-12-06
WIP kustomize

slintes:
# Description

Demonstrate (using the RT kernel MachineConfig) how to base64 encode a script and place it into a manifest template, would replace the generate and deploy script.

This based on #36!

kustomize installation: https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md

Usage:
```
cd baremetal-post
KUSTOMIZE_PLUGIN_HOME=$PWD/plugins ./kustomize build --enable_alpha_plugins ./base
```


MarSik:
@davidvossel I can see a way. We package it as a container and the admin executes it as usual (oc run...). It could contain kustomize and no event loop and end right after the execution finishes. This way we can ship updates in the usual registry way.

But we would have to figure out how to pass the login credentials (token, cert, ...) to be able to talk to openshift from inside of the container.

slintes:
> The RT kernel doesn't require any user supplied tunables. What would modifying tunables look like in this flow?

either use what kustomize provides out of the box (replacing/adding yaml values, will add an example tomorrow), or maybe another custom plugin, which e.g. reads values from a config file and applies them to mainfests.

> how would we deliver this kustomize solution in a way that allows us to push out updates, fixes

that's  indeed something which is not clear to me as well yet.

> We package it as a container 

interesting idea

> how to pass the login credentials (token...)

`kubectl run` has a `--serviceaccount` flag

slintes:
We decided to not further investigate this atm

PR #56, opened 2019-12-10
performance: generator: ensure string

fromanirh:
# Description

The KubeletConfig expects the value of "ReservedSystemCpus"
to be a string. We should ensure the manifest generator guarantees that.

Fixes # (issue)
N/A

## Type of change

## Testing

Tested using https://github.com/openshift-kni/baremetal-deploy/pull/52 - discovered during rebasing

**Test Configuration**:

- Versions: N/A
- Hardware: N/A

## Checklist

- [X] If the change is a fix, it should have an issue. The review should make sure the comments state the issue (not just the number) and it should use the keywords that will close the issue on merge.

### DIFF
--- a/features/performance/manifests/templates/12-kubeletconfig-worker-rt.yaml.in
+++ b/features/performance/manifests/templates/12-kubeletconfig-worker-rt.yaml.in
@@ -15,6 +15,6 @@ spec:
       memory: 500Mi
     cpuManagerPolicy: static
     cpuManagerReconcilePeriod: 5s
-    reservedSystemCPUs: ${RESERVED_CPUS}
+    reservedSystemCPUs: "${RESERVED_CPUS}"
     topologyManagerPolicy: best-effort
 # options: https://github.com/kubernetes/kubernetes/blob/release-1.14/pkg/kubelet/apis/config/types.go#L62


fromanirh:
The error without this patch looks like
```bash
TestPerformanceKubeletConfig
baremetal-deploy/features/performance/test/kubeletconfig_test.go:19
  CPU Manager policy
  baremetal-deploy/features/performance/test/kubeletconfig_test.go:20
    Should set the policy to 'static' [It]
    baremetal-deploy/features/performance/test/kubeletconfig_test.go:21

    Unexpected error:
        <*json.UnmarshalTypeError | 0xc000176640>: {
            Value: "number",
            Type: {
                size: 0x10,
                ptrdata: 0x8,
                hash: 3774831796,
                tflag: 7,
                align: 8,
                fieldAlign: 8,
                kind: 24,
                alg: {hash: 0x45de80, equal: 0x404280},
                gcdata: 1,
                str: 28162,
                ptrToThis: 415840,
            },
            Offset: 132,
            Struct: "KubeletConfiguration",
            Field: "reservedSystemCPUs",
        }
        json: cannot unmarshal number into Go struct field KubeletConfiguration.reservedSystemCPUs of type string
    occurred

    baremetal-deploy/features/performance/test/kubeletconfig_test.go:24
```

However, not sure my fix is the appropriate fix. I set the env like this:
```bash
export ISOLATED_CPUS=0
export RESERVED_CPUS=0
```

Maybe we should (also) document the value we expect? I believe however docs aren't enough, the generator should be robust enough to handle this known case

cynepco3hahue:
@fromanirh I checked a little bit tests, why do you not use `yaml.Unmarshal` it should cast or type correctly without the need to add quotes?

fromanirh:
> @fromanirh I checked a little bit tests, why do you not use `yaml.Unmarshal` it should cast or type correctly without the need to add quotes?

It could very much be the test is buggy. I try to use the very same code the mco operator is using. I'll doublecheck if I can just use some of its packages.

fromanirh:
And I mean it:
https://github.com/openshift/machine-config-operator/blob/master/pkg/controller/kubelet-config/helpers.go#L129
vs
https://github.com/openshift-kni/baremetal-deploy/pull/52/commits/0081abce03d42c43d6d8812230755d68bce87c86#diff-1164864c64f4bf8e09717d8c8fc7f578L42

cynepco3hahue:
Oh I believe it probably happens because of the raw config - https://github.com/openshift/machine-config-operator/blob/5bcf8d3417033734561d0174e6e821d3335ef118/pkg/apis/machineconfiguration.openshift.io/v1/types.go#L351
so It fails on decode of the raw extension

cynepco3hahue:
@fromanirh if it fixes you the problem I think we can merge it.

fromanirh:
After further testing, this seems a bug which triggers only in a narrow corner case.
The following works:
```bash
export ISOLATED_CPUS=0,1
export RESERVED_CPUS="0,1"
export RESERVED_CPUS=0-2
export ISOLATED_CPUS=0-2
```

only this doesn't:
```bash
export RESERVED_CPUS=0
export ISOLATED_CPUS=0
```

cynepco3hahue:
@fromanirh found it, so let's leave double quotes and merge it as it is now
```
Another use case for quotes is when you have a string that would be resolved as a special type. This highly depends on the YAML version and on the Schema in use. Here are some examples where you need quotes:

true, false
23
1e3
3.14159
null
```

fromanirh:
@cynepco3hahue Ack, we need LGTMs now :)

PR #83, opened 2019-12-16
Add dockerfile for the OCP CI build image

fedepaol:
# Description

Openshift CI allows to check out our repo in a src image, optionally provided by a
Dockerfile.
We have an issue right now (described here https://github.com/openshift/release/pull/6395 ) because the oc command shipped with the image provided by the ocp CI does not support applying all the files in a given folder.
With this PR we provide a dockerfile that can be used to produce an image in place of the one provided by the CI.

Fixes # (issue)

## Type of change

- [c ] Bug fix (non-breaking change which fixes an issue)

## Testing

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

- [ x ] I have performed a self-review of my own code
- [ x ] PRs should not be merged unless positively reviewed.

### DIFF
--- a/features/Dockefile.buildimage
+++ b/features/Dockefile.buildimage
@@ -1,0 +1,27 @@
+FROM fedora:31
+ENV GOPATH /go
+ENV GOBIN /go/bin
+ENV GOCACHE /go/.cache
+ENV PATH=$PATH:$GOPATH/bin
+ENV GOVERSION golang-1.13.5-1.fc31
+
+# rpms required for building and running test suites
+RUN dnf -y install \
+    $GOVERSION \
+    make \
+    tar \
+    && dnf clean all
+
+# get required golang tools and OC client
+RUN go get github.com/onsi/ginkgo/ginkgo && \
+    go get github.com/onsi/gomega/... && \
+    go get -u golang.org/x/lint/golint && \
+    export latest_oc_client_version=$(curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/ 2>/dev/null | grep -o \"openshift-client-linux-4.*tar.gz\" | tr -d \") && \
+    curl -JL https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/${latest_oc_client_version} -o oc.tar.gz && \
+    tar -xzvf oc.tar.gz && \
+    mv oc /usr/local/bin/oc && \
+    rm -f oc.tar.gz
+
+WORKDIR /src/
+
+ENTRYPOINT [ "/bin/bash" ]


features/Dockefile.buildimage
```diff
@@ -0,0 +1,25 @@
+FROM fedora:30 as build-tools
```
e-minguez: Any particular reason of using fedora:30 instead fedora:31? also the `as build-tools` I think it is not needed as it is not a multistage build (there is no FROM build-tools later)

fedepaol: None, a part the fact that I "took inspiration from" a Dockerfile built like that.
Fixing.

features/Dockefile.buildimage
```diff
@@ -0,0 +1,25 @@
...
+ENV GOPATH /go
+ENV GOBIN /go/bin
+ENV GOCACHE /go/.cache
+ENV PATH=$PATH:$GOPATH/bin
+
+# rpms required for building and running test suites
+RUN dnf -y install \
+    golang-1.12.13-1.fc30 \
```
e-minguez: is this golang specific version needed?
If so, what about having an environment variable instead?

fedepaol: Nope, but on the long term we want to use the same Dockerfile for local builds.
By pinning a specific version we make sure it won't change over time and we get reproducible builds.

Ok for the env variable.

features/Dockefile.buildimage
fedepaol: Done

features/Dockefile.buildimage
fedepaol: Done

features/Dockefile.buildimage
```diff
@@ -0,0 +1,26 @@
...
+    && dnf clean all
+
+# get required golang tools and OC client
+RUN go get github.com/onsi/ginkgo/ginkgo && \
+    go get github.com/onsi/gomega/... && \
+    go get -u golang.org/x/lint/golint && \
+    export latest_oc_client_version=$(curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/ 2>/dev/null | grep -o \"openshift-client-linux-4.*tar.gz\" | tr -d \") && \
+    curl -JL https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/${latest_oc_client_version} -o oc.tar.gz && \
+    tar -xzvf oc.tar.gz && \
+    mv oc /usr/bin/oc
```
e-minguez: nit: non packaged binaries should go into /usr/local/bin/ (for system wide binaries) or /usr/local/sbin (for root only binaries)

e-minguez: Also, removing the oc.tar.gz would be nice:

```
mv oc /usr/local/bin/oc && \
rm -f oc.tar.gz
```

fedepaol: Will do, thanks!

fedepaol: Done

fedepaol:
@e-minguez you need also to put a lgtm label

PR #96, opened 2019-12-18
Fix pre boot script and perfomance tests

yanirq:
# Description
- Fix pre boot script to take non isolated cpus value from enviroment variable.
- Fix 2e2 tests

## Type of change

- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/features/functests/performance/performance.go
+++ b/features/functests/performance/performance.go
@@ -29,7 +29,7 @@ var _ = Describe("performance", func() {
 	var _ = Context("RT Kernel Arguments", func() {
 		It("Nodes should contain kernel arguments set by machine configuration", func() {
 			nodes, err := clients.K8s.CoreV1().Nodes().List(metav1.ListOptions{
-				LabelSelector: "node-role.kubernetes.io/worker=",
+				LabelSelector: "node-role.kubernetes.io/worker-rt=",
 			})
 			Expect(err).ToNot(HaveOccurred())
 			for _, node := range nodes.Items {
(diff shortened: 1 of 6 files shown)


fedepaol:
I think this would still suffer of the problem that we don't have any node labelled as worker-rt under CI.
Would it make sense to provide a setup script that labels all the workers / half of the workers as worker-rt s?

yanirq:
> I think this would still suffer of the problem that we don't have any node labelled as worker-rt under CI.
> Would it make sense to provide a setup script that labels all the workers / half of the workers as worker-rt s?

It would make sense. not necessarily in this patch.  Tests will not fail in the current format if no node is labelled with worker-rt (The assertions are made only on nodes with that label)

fedepaol:
> > I think this would still suffer of the problem that we don't have any node labelled as worker-rt under CI.
> > Would it make sense to provide a setup script that labels all the workers / half of the workers as worker-rt s?
> 
> It would make sense. not necessarily in this patch. Tests will not fail in the current format if no node is labelled with worker-rt (The assertions are made only on nodes with that label)

Not sure I like it this way. If something in the deploy fails for some reason, and we don't have any node labelled as worker-rt, the test will skip without letting us know about the issue. If I am running these tests, it's because I know the cluster can execute them.

Also, if we don't do that the deploy phase will fail because here https://github.com/openshift-kni/baremetal-deploy/blob/master/features/performance/deploy.sh#L36 we wait for updating which won't happen since we don't have any worker-rts to update.

features/functests/performance/performance.go
```diff
@@ -29,7 +29,7 @@ var _ = Describe("performance", func() {
 	var _ = Context("RT Kernel Arguments", func() {
 		It("Nodes should contain kernel arguments set by machine configuration", func() {
 			nodes, err := clients.K8s.CoreV1().Nodes().List(metav1.ListOptions{
-				LabelSelector: "node-role.kubernetes.io/worker=",
+				LabelSelector: "node-role.kubernetes.io/worker-rt=",
```
fedepaol: Can you also assert that len(nodes) > 0? Not sure if `List` returns an error if the list is empty

features/functests/performance/performance.go
```diff
@@ -49,15 +50,16 @@ var _ = Describe("performance", func() {
 	var _ = Context("Pre boot tuning setup", func() {
 		It("Should contain a custome initrd image in boot loader", func() {
 			nodes, err := clients.K8s.CoreV1().Nodes().List(metav1.ListOptions{
-				LabelSelector: "node-role.kubernetes.io/worker=",
+				LabelSelector: "node-role.kubernetes.io/worker-rt=",
```
fedepaol: Same as above.

PR #111, opened 2019-12-20
This commit will do the following during bridge configuration in the …

### LINKED ISSUE #108: install-steps.md Drops connection when configuring bridges via ssh.
When configuring the bridges on the provision node via ssh, the connection will drop when removing the interface the connection is on. This can prevent all the commands from running if copy/pasted and prevents the ability to finish the configuration.

This can be fixed by wrapping the commands using nohup.
example: nohup bash -c 'nmcli commands'
Using nohup, once the configuration is done, the connection should resume.

Also, the steps will create a temporary connection called "Wired connection ..." when the interface is removed if the interface is active when it is removed.
Need to bring the interfaces down before removing them.

jobbler:
…provision node if applied:

    Prevent the creation of a temporary NM connection called "Wired ..." when creating bridges.    Allow the completion of the bridge configuration even if the connection drops by wrapping the commands in nohup.

# Description

Added nohup wrapper around nmcli commands that configure the bridges on the provisioning node.
Added commands to bring the interfaces down before removing them to prevent temporary "Wired ..." connections from being created in NetworkManager.

Fixes #108 install-steps.md Drops connection when configuring bridges via ssh

## Type of change

## Testing

- [x] Ran nohup'd nmcli commands as documented.

**Test Configuration**:

- Versions: 4.3.0-0.nightly-2019-12-20-082136
- Hardware: Baremetal

## Checklist

### DIFF
--- a/install-steps.md
+++ b/install-steps.md
@@ -424,18 +424,23 @@ The following steps need to be performed in order to prepare the environment.
     # You will be disconnected, but reconnect via your ssh session after running
     export PUB_CONN=<baremetal_nic_name>
     export PROV_CONN=<prov_nic_name>
-    nmcli con delete "$PROV_CONN"
-    nmcli con delete "$PUB_CONN"
-    # RHEL 8.1 appends the word "System" in front of the connection, delete in case it exists
-    nmcli con delete "System $PUB_CONN"
-    nmcli connection add ifname provisioning type bridge con-name provisioning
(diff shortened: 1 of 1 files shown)


karmab:
is this still needed ?

jobbler:
unless the baremetal-prep.sh script is deprecated, then this would still fix disconnect issues with the script. I feel it is still needed unless the script is deprecated.

rlopez133:
This is already in the install-steps.md so closing.

PR #115, opened 2019-12-22
Avoid double reboot under MCP scripts

cynepco3hahue:
# Description

This PR should merge pre-boot and RT-kernel machine config pools to avoid double reboot under asset scripts.

Fixes # (issue)

## Type of change

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist


features/performance/assets/pre-boot-tuning.sh
```diff
@@ -128,5 +119,5 @@ else
     #rpm-ostree initramfs --enable --arg=-I --arg=/etc/systemd/system.conf
     #rpm-ostree initramfs --enable --arg=-I --arg=/etc/sysconfig/irqbalance
     
-    systemctl reboot
+    touch /var/reboot
```
yanirq: What is the reason to use that here ?

cynepco3hahue: the reboot systemd unit initiates the reboot only when this file is existing

features/performance/deploy.sh
```diff
@@ -13,30 +13,52 @@ fi
...
 # W/A for https://bugzilla.redhat.com/show_bug.cgi?id=1777150
 # Apply 'worker=""' label only if not set already.
-MCPWORKER=$(oc get mcp --selector=worker="" --no-headers)
-if [ "${MCPWORKER}" == "" ] ; then
-    oc label machineconfigpool/worker worker=
-else
-    echo "MCP/worker already has label 'worker='"
-fi
+until oc label --overwrite machineconfigpool/worker worker=; do
+    sleep 5
```
yanirq: can we find ourselves stuck here in some scenarios ?  or is it ok to keep looping endlessly ? (also applies to the functions below)

cynepco3hahue: we can be stuck here only in case when something really bad happened with the environment, but at least from my experience it very rare case and the job has a global timeout.

PR #118, opened 2019-12-24
Adding performance functional tests

dorzheh:
# Description

Adding more performance tests to the E2E performance test suite

Fixes # (issue)

## Type of change

## Testing

- [x] cd <repo_name>/features and execute FEATURES=performance make functests

**Test Configuration**:

- Versions: 4.3.0-0.nightly-2019-12-13-180405
- Hardware: baremetal

## Checklist


features/functests/performance/performance.go
```diff
@@ -12,36 +13,47 @@ import (
...
+	ocv1 "github.com/openshift/api/config/v1"
 	k8sv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const testNamespace = "perfomancetest"
+const (
+	perfTestNamespace            = "perfomancetest"
+	perfWorkerNodesLabel         = "node-role.kubernetes.io/worker-rt="
+	perfWorkerRtKernelMcTemplate = "11-worker-rt-kernel"
```
yanirq: youll  need to align to the changes introduced in  https://github.com/openshift-kni/baremetal-deploy/pull/115

features/functests/performance/performance.go
```diff
@@ -50,20 +62,160 @@ var _ = Describe("performance", func() {
...
 					"grep", "-R", "initrd", "/rootfs/boot/loader/entries/").CombinedOutput()
 				Expect(err).ToNot(HaveOccurred())
-				Expect(strings.Contains(string(bootLoaderEntries), "iso_initrd.img")).To(BeTrue())
+				Expect(strings.Contains(string(bootLoaderEntries), "iso_initrd.img")).To(BeTrue(), "cannot find iso_initrd.img entry among the bootloader entries")
+			}
+		})
+	})
+
+	var _ = Context("Network latency parameters", func() {
+		It("Should contain configuration injected through the 00-tuned-network-latency template", func() {
```
yanirq: We are dependent on the tuned-network-latency profile here, for this repo it might be ok.
@cynepco3hahue  any thoughts here ?

features/functests/performance/performance.go
```diff
@@ -78,7 +230,7 @@ func mcdForNode(node *k8sv1.Node) (*k8sv1.Pod, error) {
 		return nil, err
 	}
 	// there should be one machine config deamon per node
```
yanirq: remove the comment

yanirq: remove the comment here if its in the failure description

features/functests/performance/performance.go
cynepco3hahue: Our worker-rt tuned profile depends on it, so I think it should be ok

yanirq:
@dorzheh  I would like to see the prow job running on this PR for performance tests before merging it, Can you edit one of the files under feature/performance/*  with a change  just to see the tests run?  and then revert that change ?

dorzheh:
> @dorzheh I would like to see the prow job running on this PR for performance tests before merging it, Can you edit one of the files under feature/performance/* with a change just to see the tests run? and then revert that change ? (once we see the e2e tests pass)

@yanirq , pretty weird to me that this is the only way to trigger the job but added a dummy file that will be removed later on.

yanirq:
> @yanirq , pretty weird to me that this is the only way to trigger the job but added a dummy file that will be removed later on.

This should be fixed in the ci job (openshift/release)  ill update it for future PRs

dorzheh:
@yanirq ,the previous commit solved the following issue: it seems 4.3 uses tuned 2.11 and it has an issue with the latency parsing (for example _set_latency expects an integer but receives string that cannot be "casted" to int).  According to the tuned code its fixed in v2.13+ . Hence i commented force_latency in the template. Afterwards, the tests have passed except the one checking "rt" suffix in the kernel name. Now i see the tests , that previously passed successfully , failed again. For me it  looks like the following happened: when tuning templates applied , tuned polling the changes within appropriate amount of time. It seems that tests was running before the changes were injected by the tune daemon.

features/functests/performance/performance.go
```diff
@@ -104,19 +105,15 @@ var _ = Describe("performance", func() {
...
-					param := strings.TrimSpace(line[0])
-					value := strings.TrimSpace(strings.Join(line[1:], ""))
-					if expected, ok := sysctlMap[param]; ok {
-						By(fmt.Sprintf("checking whether parameter %s value is %s", param, expected))
-						Expect(value).To(Equal(expected), fmt.Sprintf("parameter %s value is not %s", param, expected))
-					}
+				for param, expected := range sysctlMap {
+					By(fmt.Sprintf("executing the command \"sysctl -n %s\" inside the pod %s", param, mcdName))
+					Eventually(func() string {
+						out, err := exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace,
```
yanirq: looks like the new format of the command is now failing and getting an error upon execution

dorzheh:
No,the parameters doesn't exist hence the sysctl command fails

features/functests/performance/performance.go
```diff
@@ -50,20 +63,145 @@ var _ = Describe("performance", func() {
...
+				mcdName := mcd.ObjectMeta.Name
+				By(fmt.Sprintf("executing the command \"sysctl -n %s\" inside the pod %s", perfSysctlKernelVersionParam, mcdName))
+				Eventually(func() bool {
+					out, err := exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace,
+						"-c", perfMachineConfigDaemonContainer, mcdName, "sysctl", "-n", perfSysctlKernelVersionParam).CombinedOutput()
+					Expect(err).ToNot(HaveOccurred())
+					line := strings.TrimSpace(string(out))
+					return strings.Contains(line, perfSysctlKernelVersionValuePreemptEntry) &&
+						strings.Contains(line, perfSysctlKernelVersionValueRtEntry)
+				}, perfSysctlTimeout*time.Second, perfSysctlPollInterval*time.Second, ).Should(BeTrue(), "RT kernel is not installed")
```
yanirq: The last "," seems redundant or wrong .
In addition the async calls inside a loop using eventually might cause faulty behaviour here.

dorzheh:
@Yanir Quinn <yquinn@redhat.com> , the comma will be removed , though it
doesn't  affect . Regarding async calls. Indeed , it works asynchronously
and  sounds like it could be an issue if the test runs inside a  runtime
environment with limited amount of resources. However the loop is pretty
short and apparently each iteration is blocked until a result will be
received from the function.   In any case, everything works like a charm on
my systems and seems the issue is only when it runs by the CI system.

yanirq:
@dorzheh  Perhaps we should add a check for successful deployment as we do for the machineconfigs in https://github.com/openshift-kni/baremetal-deploy/blob/d7a57a0f2f948d1bdecf9b90db238049c7cdc813/features/performance/deploy.sh#L59 

Not necessarily the same way, maybe we can use Node tuning operator for that.

dorzheh:
@Yanir Quinn <yquinn@redhat.com> , do we have an opportunity to execute CI
jobs within a private repo?

yanirq:
> @yanir Quinn <yquinn@redhat.com> , do we have an opportunity to execute CI jobs within a private repo?

@fedepaol ?

features/functests/performance/performance.go
```diff
@@ -50,20 +63,145 @@ var _ = Describe("performance", func() {
...
+
+			nodes, err := clients.K8s.CoreV1().Nodes().List(metav1.ListOptions{
+				LabelSelector: perfWorkerNodesLabel,
+			})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(nodes.Items)).Should(BeNumerically(">", 0), "cannot find nodes labeled as "+perfWorkerNodesLabel)
+
+			for _, node := range nodes.Items {
+				mcd, err := mcdForNode(&node)
+				Expect(err).ToNot(HaveOccurred())
```
yanirq: can you return false here instead of Expect(err).ToNot ?

features/functests/performance/performance.go
```diff
@@ -50,20 +63,145 @@ var _ = Describe("performance", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 				mcdName := mcd.ObjectMeta.Name
-				bootLoaderEntries, err := exec.Command("oc", "rsh", "-n", "openshift-machine-config-operator", mcdName,
-					"grep", "-R", "initrd", "/rootfs/boot/loader/entries/").CombinedOutput()
+				for param, expected := range sysctlMap {
+					By(fmt.Sprintf("executing the command \"sysctl -n %s\" inside the pod %s", param, mcdName))
+					Eventually(func() string {
+						out, err := exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace,
+							"-c", perfMachineConfigDaemonContainer, mcdName, "sysctl", "-n", param).CombinedOutput()
+						Expect(err).ToNot(HaveOccurred())
```
yanirq: can you return false here instead of Expect(err).ToNot ?

features/functests/performance/performance.go
```diff
@@ -50,20 +63,145 @@ var _ = Describe("performance", func() {
...
+			for _, node := range nodes.Items {
+				mcd, err := mcdForNode(&node)
 				Expect(err).ToNot(HaveOccurred())
-				Expect(strings.Contains(string(bootLoaderEntries), "iso_initrd.img")).To(BeTrue())
+				mcdName := mcd.ObjectMeta.Name
+				By(fmt.Sprintf("executing the command \"sysctl -n %s\" inside the pod %s", perfSysctlKernelVersionParam, mcdName))
+				Eventually(func() bool {
+					out, err := exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace,
+						"-c", perfMachineConfigDaemonContainer, mcdName, "sysctl", "-n", perfSysctlKernelVersionParam).CombinedOutput()
+					Expect(err).ToNot(HaveOccurred())
```
yanirq: can you return false here instead of Expect(err).ToNot ?

features/functests/performance/performance.go
slintes: no, not here, a few lines below in the `Eventually` block. There you need to return a string which does not fullfill the expectation, in order to rerun the oc command. `Expect(err).ToNot(HaveOccurred())` does fail the test and not retry.

features/functests/performance/performance.go
```diff
@@ -50,20 +63,145 @@ var _ = Describe("performance", func() {
...
+			for _, node := range nodes.Items {
+				mcd, err := mcdForNode(&node)
+				Expect(err).ToNot(HaveOccurred())
+				mcdName := mcd.ObjectMeta.Name
+				for param, expected := range sysctlMap {
+					By(fmt.Sprintf("executing the command \"sysctl -n %s\" inside the pod %s", param, mcdName))
+					Eventually(func() string {
+						out, err := exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace,
+							"-c", perfMachineConfigDaemonContainer, mcdName, "sysctl", "-n", param).CombinedOutput()
+						Expect(err).ToNot(HaveOccurred())
```
slintes: here

features/functests/performance/performance.go
slintes: You need to return a string here which does not fulfill the expected string

dorzheh: @slintes , i think its better to remove this line (Expect..). In this case we will get more profound output, for example:
    parameter net.core.busy_poll value is not 50
    Expected
        <string>: sysctl: cannot stat /proc/sys/net/core/busy_poll: No such file or directory\ncommand terminated with exit code 255
    to equal
        <string>: 50

yanirq:
@dorzheh  the way our deployment works (as demonstrated in CI) is after applying all the machine configs and tuning profiles it performs a reboot to the node so some aspects of the real-time profiles might not be updated yet. 
We can either enter a sleep/delay at the beginning of the tests or even better check mcdForNode for running state

yanirq:
@dorzheh please rebase the PR on master

features/functests/performance/performance.go
```diff
@@ -1,73 +1,214 @@
...
+			}
+		})
+	})
+
+	// 12-worker-rt-kargs.yaml related verifications
+	var _ = Context("RT kernel arguments", func() {
+		const perfWorkerRtKernelArgsMc = "12-worker-rt-kargs"
+		It("Should contain Machine Configuration profile "+perfWorkerRtKernelArgsMc, func() {
+			_, err := clients.MachineConfig.MachineconfigurationV1().MachineConfigs().Get(perfWorkerRtKernelArgsMc, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("doesn't contain %s Machine Configiguration profile", perfWorkerRtKernelArgsMc))
```
yanirq: typo : "Configiguration"

features/functests/performance/performance.go
```diff
@@ -1,73 +1,214 @@
...
-	var _ = Context("RT Kernel Arguments", func() {
-		It("Nodes should contain kernel arguments set by machine configuration", func() {
-			nodes, err := clients.K8s.CoreV1().Nodes().List(metav1.ListOptions{
-				LabelSelector: "node-role.kubernetes.io/worker-rt=",
-			})
-			Expect(err).ToNot(HaveOccurred())
-			for _, node := range nodes.Items {
+	// 11-pre-boot-worker-tuning.yaml related verification
+	var _ = Context("Pre boot tuning adjusted by the Machine Config Operator ", func() {
+		const perfWorkerPreBootTuning = "11-worker-pre-boot-tuning"
```
yanirq: where did "11-worker-pre-boot-tuning" came from ?  did you mean "11-worker-rt-kernel" ?

yanirq: notice that both rt kernel and pre boot scripts are set on the same machine config

features/functests/performance/performance.go
```diff
@@ -1,73 +1,214 @@
...
+			checkWorkerRtProfileReadiness(nodes)
+			for _, node := range nodes {
 				mcd, err := mcdForNode(&node)
 				Expect(err).ToNot(HaveOccurred())
 				mcdName := mcd.ObjectMeta.Name
-				kargsBytes, err := exec.Command("oc", "rsh", "-n", "openshift-machine-config-operator", mcdName,
+				By("executing the command \"uname -v\" inside the pod " + mcdName)
+				Eventually(func() bool {
+					out, err := exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace,
+						"-c", perfMachineConfigDaemonContainer, mcdName, "uname", "-v").CombinedOutput()
```
yanirq: is there a chance we could miss the PREEMPT and RT values for using -v instead of -a ?
(for uname,  -a will cover everything)

features/functests/performance/performance.go
```diff
@@ -1,73 +1,214 @@
...
+					return strings.Contains(line, perfKernelVersionValuePreemptEntry) &&
+						strings.Contains(line, perfKernelVersionValueRTEntry)
+				}, perfSysctlTimeout*time.Second, perfSysctlPollInterval*time.Second).Should(BeTrue(), "RT kernel is not installed")
+			}
+		})
+	})
+
+	// 12-worker-rt-kargs.yaml related verifications
+	var _ = Context("RT kernel arguments", func() {
+		const perfWorkerRtKernelArgsMc = "12-worker-rt-kargs"
```
yanirq: this was changes to 10-worker-rt-kargs

features/functests/performance/performance.go
```diff
@@ -1,73 +1,214 @@
...
+				"net.ipv4.tcp_fastopen":           "3",
+				"kernel.numa_balancing":           "0",
+				"kernel.sched_min_granularity_ns": "10000000",
+				"vm.dirty_ratio":                  "10",
+				"vm.dirty_background_ratio":       "3",
+				"vm.swappiness":                   "10",
+				"kernel.sched_migration_cost_ns":  "5000000",
+			}
+			const perfNodeNetworkLatencyTuned = "openshift-node-network-latency"
+			_, err := clients.CntoConfig.TunedV1().Tuneds(perfClusterNodeTuningOperatorNamespaces).Get(perfNodeNetworkLatencyTuned,metav1.GetOptions{})
```
yanirq: im not a big fan of adding a lot of dependencies for this one liner (if this is the case) if the check turns out to be un-needed then lets remove it.

features/functests/performance/performance.go
dorzheh: @yanirq , no uname -v works fine in our case

features/functests/performance/performance.go
dorzheh: @yanirq , ah , forgot to rebase my stuff.

features/functests/performance/performance.go
dorzheh: @yanirq . Thanks

features/functests/performance/performance.go
dorzheh: @yanirq , perhaps. The other way  - execute the oc command and parse the output. Lets the CI to complete all the tests successfully and afterwards decide about more convenient solution.

features/functests/performance/performance.go
```diff
@@ -254,12 +229,12 @@ func checkWorkerRtProfileReadiness(nodes []k8sv1.Node) {
...
+func checkFileExistence(nodes []k8sv1.Node, file string) {
 	for _, node := range nodes {
 		mcd, err := mcdForNode(&node)
 		Expect(err).ToNot(HaveOccurred())
 		mcdName := mcd.ObjectMeta.Name
-		By(fmt.Sprintf("executing the command \"ls %s\" inside the pod %s", file, mcdName))
+		By(fmt.Sprintf("Searching for the file %s.Executing the command \"ls %s\" inside the pod %s", file, file, mcdName))
 		err = exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace, "-c",
 			perfMachineConfigDaemonContainer, mcdName, "ls", "/rootfs/"+file).Run()
 		Expect(err).To(BeNil(), "cannot find the script "+file)
```
yanirq: why are we checking err=nil here ?

yanirq: why are we checking err=nil here ?

features/functests/performance/performance.go
```diff
@@ -77,17 +218,51 @@ func mcdForNode(node *k8sv1.Node) (*k8sv1.Pod, error) {
...
+// Check whether appropriate file exists on the system
+func checkFileExistense(nodes []k8sv1.Node, file string) {
+	for _, node := range nodes {
+		mcd, err := mcdForNode(&node)
+		Expect(err).ToNot(HaveOccurred())
+		mcdName := mcd.ObjectMeta.Name
+		By(fmt.Sprintf("executing the command \"ls %s\" inside the pod %s", file, mcdName))
+		err = exec.Command("oc", "rsh", "-n", mcd.ObjectMeta.Namespace, "-c",
+			perfMachineConfigDaemonContainer, mcdName, "ls", "/rootfs/"+file).Run()
+		Expect(err).To(BeNil(), "cannot find the script "+file)
```
gsr-shanks: Shouldn't the err read as "cannot find the file " instead of "cannot find the script "?

install-steps.md
```diff
@@ -603,17 +609,17 @@ to fix the ConfigMap for the Metal3 operator. This ConfigMap is used to notify
...
       cache_url: ''
-      deploy_kernel_url: http://172.22.0.3:6180/images/ironic-python-agent.kernel
-      deploy_ramdisk_url: http://172.22.0.3:6180/images/ironic-python-agent.initramfs
+      deploy_kernel_url: http://<provisioning_ip>:6180/images/ironic-python-agent.kernel
+      deploy_ramdisk_url: http://<provisioning_ip>:6180/images/ironic-python-agent.initramfs
       dhcp_range: 172.22.0.10,172.22.0.100
       http_port: "6180"
-      ironic_endpoint: http://172.22.0.3:6385/v1/
+      ironic_endpoint: http://<provisioning_ip>:6385/v1/
       ironic_inspector_endpoint: http://172.22.0.3:5050/v1/
```
gsr-shanks: Maybe we would want to remove the above IP as well.

yanirq: @gsr-shanks @dorzheh this looks like it doesn't belong to this PR. Another rebase against the master branch should align the files.

dorzheh:
Closing this PR. Suppressed by #124

PR #122, opened 2020-01-08
Added 3 more tests to sctp feature. Tests: KernelModule is loaded, co…

nikita133:
…nnectivity via Servis and NodeIP

# Description

Fixes # (issue)

## Type of change

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/features/functests/sctp/sctp.go
+++ b/features/functests/sctp/sctp.go
@@ -1,6 +1,8 @@
 package sctp
 
 import (
+	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"time"
@@ -16,6 +18,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/pointer"
 )
 
 const mcYaml = "../sctp/sctp_module_mc.yaml"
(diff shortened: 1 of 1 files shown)


features/functests/sctp/sctp.go
```diff
@@ -36,56 +38,104 @@ var _ = Describe("sctp", func() {
...
+				}, 3*time.Minute, 1*time.Second).Should(Equal(k8sv1.PodRunning))
 
-			By("Connecting a client to the server")
-			clientArgs := []string{"-ip", runningPod.Status.PodIP, "-port", "30101", "-lport", "30102"}
-			clientPod := scptTestPod("sctpclient", clientNode, "sctpclient", clientArgs)
-			clients.K8s.CoreV1().Pods(testNamespace).Create(clientPod)
+			})
+			Context("Client Server Connection", func() {
+				// OCP-26760
+				FIt("Should connect a client pod to a server pod", func() {
```
fedepaol: remove focus

features/functests/sctp/sctp.go
```diff
@@ -36,56 +38,104 @@ var _ = Describe("sctp", func() {
...
-					Name:          "sctpport",
-					Protocol:      k8sv1.ProtocolSCTP,
-					ContainerPort: 30101,
-				},
-			}
-			serverPod, err := clients.K8s.CoreV1().Pods(testNamespace).Create(pod)
-			Expect(err).ToNot(HaveOccurred())
+		// OCP-26759
+		Context("Test Kerenel Module", func() {
+			FIt("Kernel Module is loaded", func() {
```
fedepaol: Remove Focus

features/functests/sctp/sctp.go
```diff
@@ -36,56 +38,104 @@ var _ = Describe("sctp", func() {
...
-				return pod.Status.Phase
-			}, 1*time.Minute, 1*time.Second).Should(Equal(k8sv1.PodSucceeded))
+					Eventually(func() k8sv1.PodPhase {
+						pod, err := clients.K8s.CoreV1().Pods(testNamespace).Get(serverPod.Name, metav1.GetOptions{})
+						Expect(err).ToNot(HaveOccurred())
+						return pod.Status.Phase
+					}, 1*time.Minute, 1*time.Second).Should(Equal(k8sv1.PodSucceeded))
+				})
+				// OCP-26761
+				FIt("Should connect a client pod to a server pod via Service ClusterIP", func() {
```
fedepaol: remove focus

features/functests/sctp/sctp.go
```diff
@@ -104,7 +154,7 @@ func loadMC() *mcfgv1.MachineConfig {
 
 func checkForSctpReady(client *kubernetes.Clientset) {
 	nodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{
-		LabelSelector: "node-role.kubernetes.io/worker=",
+		LabelSelector: "node-role.kubernetes.io/worker-sctp=",
```
fedepaol: We may end up with a worker-A / worker-B machine pools, so this label is likely to change.

features/functests/sctp/sctp.go
```diff
@@ -36,56 +38,104 @@ var _ = Describe("sctp", func() {
...
 
-			By("Fetching the server's ip address")
-			var runningPod *k8sv1.Pod
-			Eventually(func() k8sv1.PodPhase {
-				runningPod, err = clients.K8s.CoreV1().Pods(testNamespace).Get(serverPod.Name, metav1.GetOptions{})
+		var _ = Describe("Test Connectivity", func() {
+			BeforeEach(func() {
+				By("Choosing the nodes for the server and the client")
+				nodes, err = clients.K8s.CoreV1().Nodes().List(metav1.ListOptions{
+					LabelSelector: "node-role.kubernetes.io/worker-sctp=",
```
fedepaol: We may end up with a worker-A / worker-B machine pools, so this label is likely to change.

features/functests/sctp/sctp.go
```diff
@@ -36,56 +38,104 @@ var _ = Describe("sctp", func() {
...
-			clientNode = nodes.Items[0].ObjectMeta.Labels[hostnameLabel]
-			serverNode = nodes.Items[0].ObjectMeta.Labels[hostnameLabel]
-			if len(nodes.Items) > 1 {
-				serverNode = nodes.Items[1].ObjectMeta.Labels[hostnameLabel]
-			}
+		var activeService *k8sv1.Service
+		var runningPod *k8sv1.Pod
+		var serverPod *k8sv1.Pod
+		var nodes *k8sv1.NodeList
+		var err error
```
fedepaol: err can be scoped locally

features/functests/sctp/sctp.go
```diff
@@ -36,56 +38,104 @@ var _ = Describe("sctp", func() {
...
+				checkForSctpReady(clients.K8s)
+			})
+		})
 
-			By("Fetching the server's ip address")
-			var runningPod *k8sv1.Pod
-			Eventually(func() k8sv1.PodPhase {
-				runningPod, err = clients.K8s.CoreV1().Pods(testNamespace).Get(serverPod.Name, metav1.GetOptions{})
+		var _ = Describe("Test Connectivity", func() {
+			BeforeEach(func() {
```
fedepaol: This is leaking pods between tests if something goes wrong.
You can either use a beforeAll function OR delete the pods in after each

features/functests/sctp/sctp.go
nikita133: Moved to BeforeEach

features/functests/sctp/sctp.go
nikita133: Relevant code added to section AfterEach

features/functests/sctp/sctp.go
```diff
@@ -35,57 +38,143 @@ var _ = Describe("sctp", func() {
...
-			Expect(err).ToNot(HaveOccurred())
-			clientNode = nodes.Items[0].ObjectMeta.Labels[hostnameLabel]
-			serverNode = nodes.Items[0].ObjectMeta.Labels[hostnameLabel]
-			if len(nodes.Items) > 1 {
-				serverNode = nodes.Items[1].ObjectMeta.Labels[hostnameLabel]
-			}
+		var activeService *k8sv1.Service
+		var runningPod *k8sv1.Pod
+		var serverPod *k8sv1.Pod
+		var nodes *k8sv1.NodeList
```
fedepaol: It would be better to have locally scoped nodes. Here nodes is leaked among tests, but initialized differently each time.
I think it better to keep these variable as more locally scoped as we can.
For `nodes` for example, we can have two different instances, one in `Describe("Test Connectivity Negative", func()` and one other in `var _ = Describe("Test Connectivity", func() { ` scope

This is the same for the other variables. The more local they can be, the better (even if you split them).

features/functests/sctp/sctp.go
fedepaol: Let's do the other way around, labelling only some nodes as worker-sctp

features/functests/sctp/sctp.go
```diff
@@ -35,57 +38,143 @@ var _ = Describe("sctp", func() {
...
+					sctpClientServerConnection(activeService.Spec.ClusterIP,
+						activeService.Spec.Ports[0].Port, clientNode, serverPod.Name, k8sv1.PodSucceeded)
+				})
+				// OCP-26762
+				It("Should connect a client pod to a server pod via Service NodeIP", func() {
+					sctpClientServerConnection(nodes.Items[0].Status.Addresses[0].Address,
+						activeService.Spec.Ports[0].Port, clientNode, serverPod.Name, k8sv1.PodSucceeded)
+				})
+				AfterEach(func() {
+					namespace.Clean("sctptest", clients.K8s)
```
fedepaol: Can we have this in the outermost AfterEach? So it won't require duplicating it

features/functests/sctp/sctp.go
```diff
@@ -35,57 +38,143 @@ var _ = Describe("sctp", func() {
...
+				})
+			})
+			AfterEach(func() {
+				deleteService(clients.K8s)
```
fedepaol: Same here

features/functests/sctp/sctp.go
```diff
@@ -102,9 +191,24 @@ func loadMC() *mcfgv1.MachineConfig {
 	return mc
 }
 
+func featureLatencySensitiveOn() bool {
```
fedepaol: this could be done using the type safe client. I was using it before the feature gate was enabled by default.
You can check it here https://github.com/fedepaol/sctp_test/commit/ee11f006b88fc797714b228fb406b3f5c95fb573#diff-d9cedae8764d88c1b05a71fb306ee070 (on the part that was being removed because there was no need to set the feature gate anymore)

rlopez133:
@fedepaol - is this ready to merge?

fedepaol:
@rlopez133 forgot about this! Tests (including these) were moved to https://github.com/openshift-kni/cnf-features-deploy , so this is safe to close

PR #123, opened 2020-01-09
Provide a more detailed description of how enable sctp and what it is

fedepaol:
# Description

This PR is meant to describe better what the machine config does, and what we can do with the sctp protocol inside the OpenShift platform.

Fixes # (issue)

## Type of change

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/features/sctp/README.md
+++ b/features/sctp/README.md
@@ -1,7 +1,92 @@
-# How to enable SCTP on OCP 4.3
+# SCTP on OpenShift
 
-## Load the sctp kernel module
+OpenShift Container Platform includes the capability of using (single homed) SCTP connections.
+The Stream Control Transmission Protocol (SCTP) is a computer networking communications protocol which operates at the transport layer and serves a role similar to the popular protocols TCP and UDP.
 
-The SCTP module is blacklisted by default, so in order to use it we need to unblacklist it and load it at boot time.
+Within the OpenShift Container Platform you can:
 
-This is achieved by the Machine Config configuration file [sctp_module_mc.yaml](./sctp_module_mc.yaml).
+- establish single homed pod to pod sctp connections
+- expose SCTP ClusterIP Services
+- expose SCTP NodePort Services
+
+This is achieved by specifying the `protocol` field to `SCTP` the same way is done with `TCP`.
+
+For example:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: demoserver
+  namespace: sctp-demo
+  labels:
+    app: demoserver
+spec:
+  selector:
+    app: demoserver
+  ports:
+    - name: demoserver
+      protocol: SCTP
+      port: 30101
+      targetPort: "demoserver"
+```
+
+## Enabling SCTP
+
+The `SCTP` protocol is enabled by default in OpenShift. Nothing prevents the creation of Pods exposing SCTP ports.
+
+However, in order to be able to use it some tweaking is necessary. This is because the SCTP kernel module is blacklisted by default.
+
+In order for applications to be able to load the module, two different actions need to be performed:
+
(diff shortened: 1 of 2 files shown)


features/sctp/README.md
```diff
@@ -1,7 +1,92 @@
...
+          mode: 420
+          path: /etc/modprobe.d/sctp-blacklist.conf
+        - contents:
+            source: data:text/plain;charset=utf-8;base64,c2N0cA==
+          filesystem: root
+          mode: 420
+          path: /etc/modules-load.d/sctp-load.conf
+```
+
+Please note that `c2N0cA==` is the `sctp` string encoded via the base64 encoding.
```
fromanirh: I'd rather sau that (per machineconfig spec) the value of the "source" field should be data url content: https://github.com/vincent-petithory/dataurl

features/sctp/README.md
```diff
@@ -1,7 +1,92 @@
...
+    storage:
+      files:
+        - contents:
+            source: data:,
+            verification: {}
+          filesystem: root
+          mode: 420
+          path: /etc/modprobe.d/sctp-blacklist.conf
+        - contents:
+            source: data:text/plain;charset=utf-8;base64,c2N0cA==
```
yuvalk: AFAIK, since 'sctp' is within the 'safe range', it'll work without base64
and be more readable as

`source: data:text/plain;charset=utf-8;sctp`

fedepaol: Yes, I just tried it to be sure.
According to the [rfc](https://tools.ietf.org/html/rfc2397), it should be `data:text/plain;charset=utf-8,sctp` (and it worked this way).

Will update both the manifest & the doc.

features/sctp/README.md
fedepaol: Thanks for mentioning, as per the comment below I'll get rid of the base64 thing. It will end up in a more clear manifest.

features/sctp/README.md
```diff
@@ -1,7 +1,92 @@
-# How to enable SCTP on OCP 4.3
+# SCTP on OpenShift
 
-## Load the sctp kernel module
+OpenShift Container Platform includes the capability of using (single homed) SCTP connections.
+The Stream Control Transmission Protocol (SCTP) is a computer networking communications protocol which operates at the transport layer and serves a role similar to the popular protocols TCP and UDP.
 
-The SCTP module is blacklisted by default, so in order to use it we need to unblacklist it and load it at boot time.
+Withing the OpenShift Container Platform you can:
```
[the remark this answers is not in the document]
fedepaol: Done

features/sctp/README.md
```diff
@@ -1,7 +1,92 @@
...
+      targetPort: "demoserver"
+```
+
+## Enabling SCTP
+
+The `SCTP` protocol is enabled by default in OpenShift. Nothing prevents the creation of Pods exposing SCTP ports.
+
+However, in order to be able to use it some tweaking is necessary. This is because the SCTP kernel module is blacklisted by default.
+
+In order to have the applications able to load the module, two different actions need to be performed:
```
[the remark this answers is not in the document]
fedepaol: Done

features/sctp/README.md
```diff
@@ -1,7 +1,92 @@
...
+        - contents:
+            source: data:text/plain;charset=utf-8,sctp
+          filesystem: root
+          mode: 420
+          path: /etc/modules-load.d/sctp-load.conf
+```
+
+Please note that the file content is supposed to use the [data url scheme](https://tools.ietf.org/html/rfc2397).
+
+This means that the effect of applying this is to:
```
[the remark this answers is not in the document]
fedepaol: Done

Issue #137, opened 2020-02-04
Feature Enhancement of ansible-ipi-install - Clean-up old bootstrap


abays:
In the event of a failed deployment, it is possible that the bootstrap VM and volumes are left behind.  We should have the Ansible tool check for these old resources and remove them during execution of the installer tasks.

PR #138, opened 2020-02-04
Clear bootstrap resources with Ansible

abays:
# Description

If a deployment fails before the installer reaches the step where it removes the bootstrap VM and its volumes, then those resources will also be present during a re-run of the playbook.  We should make sure to clear any such stale resources before executing the installer binary.

Fixes # https://github.com/openshift-kni/baremetal-deploy/issues/137

## Type of change

## Testing

Simply execute the Ansible playbook and wait until the bootstrap VM has been created, then ctrl+C to stop execution.  Re-run the playbook and make sure the new task to remove the bootstrap VM + volumes succeeds.

**Test Configuration**:

- Versions:
- Hardware:

## Checklist


abays:
@rlopez133 Ack, will do

abays:
@rlopez133 Another good call!  I'll do that as well.

rlopez133:
> @rlopez133 Another good call! I'll do that as well.

Actually re-reading this I'm wondering 2 things
* instead of renaming my files, call it 59_* in the installer role?
OR 
* having it do this cleanup in node-prep as a check after the 80_libvirt creates the pools? then have the file called like 81_*?

abays:
> > @rlopez133 Another good call! I'll do that as well.
> 
> Actually re-reading this I'm wondering 2 things
> 
>     * instead of renaming my files, call it 59_* in the installer role?
>       OR
> 
>     * having it do this cleanup in node-prep as a check after the 80_libvirt creates the pools? then have the file called like 81_*?

Hmm.  It seems like the first option would be preferable.  My thinking there is that maybe someone would want to create separate playbooks for node-prep and installer roles, and perhaps they would be re-running the installer playbook over and over while working towards a successful deploy.  If that were the case, we'd want the installer role to then clean-up any stale bootstrap resources.  Thoughts?

rlopez133:
> > > @rlopez133 Another good call! I'll do that as well.
> > 
> > 
> > Actually re-reading this I'm wondering 2 things
> > ```
> > * instead of renaming my files, call it 59_* in the installer role?
> >   OR
> > 
> > * having it do this cleanup in node-prep as a check after the 80_libvirt creates the pools? then have the file called like 81_*?
> > ```
> 
> Hmm. It seems like the first option would be preferable. My thinking there is that maybe someone would want to create separate playbooks for node-prep and installer roles, and perhaps they would be re-running the installer playbook over and over while working towards a successful deploy. If that were the case, we'd want the installer role to then clean-up any stale bootstrap resources. Thoughts?



> > > @rlopez133 Another good call! I'll do that as well.
> > 
> > 
> > Actually re-reading this I'm wondering 2 things
> > ```
> > * instead of renaming my files, call it 59_* in the installer role?
> >   OR
> > 
> > * having it do this cleanup in node-prep as a check after the 80_libvirt creates the pools? then have the file called like 81_*?
> > ```
> 
> Hmm. It seems like the first option would be preferable. My thinking there is that maybe someone would want to create separate playbooks for node-prep and installer roles, and perhaps they would be re-running the installer playbook over and over while working towards a successful deploy. If that were the case, we'd want the installer role to then clean-up any stale bootstrap resources. Thoughts?

That works for me. Lets do the 59_* that way we can leave the original files with their original names.

abays:
@rlopez133 I see the `virt` and `virt_pool` modules in the Ansible docs, but I don't see a way to explicitly delete VM volumes.  Are you aware of a way to avoid using `cmd/shell` for that in particular?

rlopez133:
> @rlopez133 I see the `virt` and `virt_pool` modules in the Ansible docs, but I don't see a way to explicitly delete VM volumes. Are you aware of a way to avoid using `cmd/shell` for that in particular?

Looking to see. Curious, if you use virt to destroy the vm, does it not remove the volume resource as well?

abays:
> > @rlopez133 I see the `virt` and `virt_pool` modules in the Ansible docs, but I don't see a way to explicitly delete VM volumes. Are you aware of a way to avoid using `cmd/shell` for that in particular?
> 
> Looking to see. Curious, if you use virt to destroy the vm, does it not remove the volume resource as well?

Unfortunately, it doesn't seem to do so:

```[root@goblin clusterconfigs]# virsh list --all
 Id    Name                           State
----------------------------------------------------

[root@goblin clusterconfigs]# virsh vol-list default
 Name                 Path                                    
------------------------------------------------------------------------------
 goblin-l792q-bootstrap /var/lib/libvirt/images/goblin-l792q-bootstrap
 goblin-mdgtf-bootstrap /var/lib/libvirt/images/goblin-mdgtf-bootstrap
 goblin-mdgtf-bootstrap.ign /var/lib/libvirt/images/goblin-mdgtf-bootstrap.ign
 goblin-rgr6q-bootstrap /var/lib/libvirt/images/goblin-rgr6q-bootstrap
 goblin-rzz6g-bootstrap /var/lib/libvirt/images/goblin-rzz6g-bootstrap
 goblin-srrmj-bootstrap /var/lib/libvirt/images/goblin-srrmj-bootstrap
 goblin-srrmj-bootstrap.ign /var/lib/libvirt/images/goblin-srrmj-bootstrap.ign
 goblin-wbghs-bootstrap /var/lib/libvirt/images/goblin-wbghs-bootstrap```

**Comment by rlopez133** [comment]
> > > @rlopez133 I see the `virt` and `virt_pool` modules in the Ansible docs, but I don't see a way to explicitly delete VM volumes. Are you aware of a way to avoid using `cmd/shell` for that in particular?
> > 
> > 
> > Looking to see. Curious, if you use virt to destroy the vm, does it not remove the volume resource as well?
> 
> Unfortunately, it doesn't seem to do so:
> 
> ```
>  Id    Name                           State
> ----------------------------------------------------
> 
> [root@goblin clusterconfigs]# virsh vol-list default
>  Name                 Path                                    
> ------------------------------------------------------------------------------
>  goblin-l792q-bootstrap /var/lib/libvirt/images/goblin-l792q-bootstrap
>  goblin-mdgtf-bootstrap /var/lib/libvirt/images/goblin-mdgtf-bootstrap
>  goblin-mdgtf-bootstrap.ign /var/lib/libvirt/images/goblin-mdgtf-bootstrap.ign
>  goblin-rgr6q-bootstrap /var/lib/libvirt/images/goblin-rgr6q-bootstrap
>  goblin-rzz6g-bootstrap /var/lib/libvirt/images/goblin-rzz6g-bootstrap
>  goblin-srrmj-bootstrap /var/lib/libvirt/images/goblin-srrmj-bootstrap
>  goblin-srrmj-bootstrap.ign /var/lib/libvirt/images/goblin-srrmj-bootstrap.ign
>  goblin-wbghs-bootstrap /var/lib/libvirt/images/goblin-wbghs-bootstrap```
> ```

@abays Ok. What if we destroy the storage pool and re-create it every time? 

If you do
```
- virt_pool:
    command: delete
    name: default
```

^ Should destroy the storage pool and its contents (accordinging to the doc). However, then your going to need to recreate it. 

In that yml file after you do the delete, you can likely call the libvirt role to recreate it for you
```
- name: Re-creating the libvirt pool
  import_tasks: roles/node-prep/80_libvirt_pool.yml
```

and see if that works? I'd confirm that deleting the storage pool indeed does what you expect first before trying the import (which i hope is the right syntax :) )

**Comment by abays** [comment]
I was actually thinking of the same thing.  I had seen in the Ansible docs that it mentioned that deleting the pool also deletes all its contents, which hopefully means it takes care of the volumes (will test, of course :) ).  

Your approach sounds good!

**Comment by abays** [comment]
@rlopez133 Unfortunately it looks like the Ansible docs for the `virt_pool` module are mistaken.  They say:

```
# Delete a storage pool (destroys contents)
- virt_pool:
    command: delete
    name: vms
```

...but the "destroys contents" part seems to be incorrect.  With lingering volumes in the pool, I get this error:

```
TASK [installer : Stop and delete default libvirt pool to delete old bootstrap volumes] ****************************************************************************************************************************************************************************************************
Wednesday 05 February 2020  11:50:49 -0500 (0:00:00.039)       0:02:30.461 **** 
ok: [127.0.0.1] => (item=destroy)
failed: [127.0.0.1] (item=delete) => {"ansible_loop_var": "item", "changed": false, "item": "delete", "msg": "failed to remove pool '/var/lib/libvirt/images': Directory not empty"}
```

I will try messing with the module options to see if I can force the removal.  If that doesn't work, I think we might be forced to fall back to the `shell` module for the volumes.

**Comment by rlopez133** [comment]
> @rlopez133 Unfortunately it looks like the Ansible docs for the `virt_pool` module are mistaken. They say:
> 
> ```
> # Delete a storage pool (destroys contents)
> - virt_pool:
>     command: delete
>     name: vms
> ```
> 
> ...but the "destroys contents" part seems to be incorrect. With lingering volumes in the pool, I get this error:
> 
> ```
> TASK [installer : Stop and delete default libvirt pool to delete old bootstrap volumes] ****************************************************************************************************************************************************************************************************
> Wednesday 05 February 2020  11:50:49 -0500 (0:00:00.039)       0:02:30.461 **** 
> ok: [127.0.0.1] => (item=destroy)
> failed: [127.0.0.1] (item=delete) => {"ansible_loop_var": "item", "changed": false, "item": "delete", "msg": "failed to remove pool '/var/lib/libvirt/images': Directory not empty"}
> ```
> 
> I will try messing with the module options to see if I can force the removal. If that doesn't work, I think we might be forced to fall back to the `shell` module for the volumes.

@abays : See if this works

```
- name: Destroy Storage Pool for default
  virt_pool:
    command: destroy
    name: default
    xml: '{{ lookup("template", "roles/node-prep/templates/dir.xml.j2") }}'
  become: yes
```

**Comment by rlopez133** [comment]
> Looks like we don't need to use the find, file to remove the disks, lets use ovirt_disk https://docs.ansible.com/ansible/latest/modules/ovirt_disk_module.html . I can likely look at this shortly

After looking at this closer, this module requires us to setup ovirt engine and all that to do a simple task. I don't think its worth it. Not to mention that the package dependency is not in RHEL 8.1 yet.

File: ansible-ipi-install/roles/installer/tasks/59_cleanup_bootstrap.yml [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+    name: "{{ item }}"
+    command: undefine
+  with_items: "{{ all_vms.list_vms }}"
+  when: item is search("bootstrap")
+  become: yes
+
+- name: Find old bootstrap VM Storage
+  find:
+    paths: "{{ default_libvirt_pool_dir }}"
+    patterns: '*-bootstrap,*-bootstrap.ign'
```
e-minguez: I'm not sure if this is still valid but back in the day there were three volumes, those two + `*-bootstrap-orig`

ansible-ipi-install/roles/installer/tasks/59_cleanup_bootstrap.yml
```diff
@@ -0,0 +1,43 @@
...
+  register: find_results
+  become: yes
+
+- name: Delete old bootstrap VMs Storage
+  file:
```
e-minguez: I'm not sure if removing the storage is enough or if `vol-delete` needs to be used. Otherwise I think you can delete them then refreshing the pool.

ansible-ipi-install/roles/installer/tasks/59_cleanup_bootstrap.yml
abays: I haven't seen the `-bootstrap-orig` in recent builds.

ansible-ipi-install/roles/installer/tasks/59_cleanup_bootstrap.yml
abays: So the reason we're doing it this way is to use an idempotent module rather than `cmd/shell` call to `vol-delete`.  Deleting the files from the disk does in fact leave old data in `virsh vol-list`, temporarily, but this is refreshed during `60_deploy_ocp.yml` when the new bootstrap volumes are created.

PR #139, opened 2020-02-04
Shared labs

dustinblack:
# Description

Enhances the baremetal-deploy ansible-ipi-install roles to include some core functionality required for shared lab environments

1.  Moves custom manifests to a subdirectory so that we can use the role `files` directory for ignition config changes
2. Adds playbook to customize the node filesystems through the ignition config files by using a "fake" root subdirectory in the role `files` directory.
3. Adds the ability to select between `firewalld` and `iptables` firewall types

## Type of change

## Testing

- [x] Partial completion of ansible-automate deployment in a shared lab (so far up to the point where these changes target their effect, but an install has not been fully completed)

**Test Configuration**:

- Versions: latest-4.3
- Hardware: Dell r640 in a shared lab environment

## Checklist


dustinblack:
FYI @rlopez133 @mkarg75

ansible-ipi-install/README.md
```diff
@@ -316,13 +316,38 @@ Sample playbook.yml
   - installer
 ~~~
 
+## Customizing the Node Filesystems
+If you need to modify files on the node filesystems, you can augment the "fake"
+roots for the masters and workers under the 
+`roles/installer/files/customize_filesystem/{master,worker}` directories.
```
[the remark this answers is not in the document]
dustinblack: The filetranspiler tool will take a directory structure from a "fake" root starting point and merge that structure and file content into the ignition file. I can try to make this more clear in the README and point out to the filetranspiler doc as well.

ansible-ipi-install/README.md
```diff
@@ -316,13 +316,38 @@ Sample playbook.yml
...
+~~~sh
+IFCFG_DIR="roles/installer/files/customize_filesystem/master/etc/sysconfig/network-scripts"
+IFNAME="eno1"
+mkdir -p $IFCFG_DIR
+cat << EOF > $IFCFG_DIR/ifcfg-${IFNAME}
+DEVICE=${IFNAME}
+BOOTPROTO=none
+ONBOOT=no
+EOF
+~~~
```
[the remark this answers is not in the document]
dustinblack: Right. the `roles/installer/files/customize_filesystem/master` directory becomes the "fake" root for the master config, and by placing the `etc/sysconfig/network-scripts/ifcfg-eno1` file there, the file will be added to the `master.ign` file and therefore permanently change the master node filesystem structure with this augmentation.

ansible-ipi-install/README.md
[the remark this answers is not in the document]
rlopez133: That be great, to be honest, when I read it I just wasn't sure what it did or really how I needed to use it.

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -16,6 +16,10 @@ pub_nic=eno2
 # (Optional) Activation-key org_id for proper setup of subscription-manager, empty value skips registration
 #org_id=""
 
+# Firewall type to use.
+# Values accepted: 'firewalld', 'iptables' (defaults to 'firewalld')
+firewall="firewalld"
+
```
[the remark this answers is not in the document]
dustinblack: This is a lab requirement. Well, requirement is a strong word, but our shared labs are provisioned to use iptables by default. This change allows us to simply act as consumers of the lab environment as provisioned by the foreman automation with little overhead.

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
```diff
@@ -0,0 +1,23 @@
...
+    /usr/local/bin/openshift-baremetal-install --dir {{ dir }} create ignition-configs
+
+- name: Get Filetranspiler
+  get_url:
+    url: https://raw.githubusercontent.com/ashcrow/filetranspiler/master/filetranspile
+    dest: /var/tmp/filetranspile.py
+    mode: '0755'
+
+- name: Move Ignition Config Files
+  command: mv {{ dir }}/{{ item }}.ign {{ dir }}/{{ item }}.ign.orig
```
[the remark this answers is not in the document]
dustinblack: fair enough

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
```diff
@@ -0,0 +1,23 @@
...
+  with_items:
+    - "master"
+    - "worker"
+
+- name: Augment Ignition Config Files
+  shell:
+    /var/tmp/filetranspile.py -i {{ dir }}/{{ item }}.ign.orig -f roles/installer/files/customize_filesystem/{{ item }} -o {{ dir }}/{{ item }}.ign
```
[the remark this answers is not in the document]
dustinblack: I've had my hand slapped for using `/tmp` for similar situations in the past. I can't exactly recall why, but it worked and now I have a new habit. Is there a particular argument to use `/tmp` instead here?

I could also just put it in `/home/kni` I guess.

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -2,6 +2,7 @@
 # tasks file for installer
 - include_tasks: 10_get_oc.yml 
 - include_tasks: 20_extract_installer.yml
+- include_tasks: 25_customize_filesystem.yml
```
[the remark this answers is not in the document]
dustinblack: Correct. I thought about adding some test logic to decide if we should run this, but I put on my ansible hat and tried to think and design idempotently instead of overloading with script logic. If the "fake" root directories are empty, then the filetranspile task will simply result in outputting unchanged ignition config files.

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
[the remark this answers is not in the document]
rlopez133: I guess because `/tmp` can be cleaned up. Yeah I'd prefer {{ ansible_user_dir }} as you know that will exist.

dustinblack:
> Unsure, but do you know if we'd still need the pkill dhclient? Reason I bring this up is your public_nic would of gotten a dhcp address using dhclient at some pt.

I'm not 100% sure on this. The existing `ifup / ifdown` commands I found to cause a problem with the libvirt bridge in my environment, leading to a situation where everything appears fine, but the VM won't get a DHCP address on the baremetal interface. This I'm sure is particular to our lab environment where we need to run our own dnsmasq on the deployment node to serve the baremetal network, so that bridge already exists before the ansible tasks try to create it. In my experience, NM and libvirt can get out-of-sync with each other (for lack of a better description -- I'm sure it's a bug somewhere in the stacks), and I need to use `nmcli` commands to make sure things get back to behaving.

What _should_ happen with the `nmcli reload` and `up` commands is that the NM-controlled dhclient processes will restart, so having the same effect as the manual process kill.

dustinblack:
@rlopez133 Fixes we discussed are in place, but let me re-test in my env first...

ansible-ipi-install/README.md
```diff
@@ -318,13 +318,47 @@ Sample playbook.yml
...
+
+NOTE: Do not place any files directly in the "fake" root -- only in subdirectories.
+Files in the root will cause the ignition process to fail. (There is a task in the 
+playbook to cleanup the `.gitkeep` file in the root, if it is left in place.)
+
+This will utilize the Ignition 
+[filetranspiler tool](https://github.com/ashcrow/filetranspiler/blob/master/filetranspile), 
+which you can read about for more information on how to use the "fake" root directories.
+
+An example of using this customization is to disable a network interface that
```
abays: Will this example filesystem change also require MCO manifests (to be include as directed in `roles/installer/files/manifests`) for post-first-boot consistency?  When we ran into this problem using UPI last year (wanting to disable an external interface), we found that we had to change the ignition and add manifests for the MCO.  Otherwise the MCO would take issue with the filesystem being in an unexpected state and the deployment would fail.

dustinblack: I haven't experienced this, but that's not to say it won't happen. I can confirm that this process, as defined here, will allow me to get master nodes deployed with the Ansible-IPI automation, and I can do this same ignition config trick with a manual UPI method and have it work for all master and worker nodes. So if you are saying that the actual deployment will not complete without the manifest changes, I don't believe that affects us.

However, our clusters have so far been short-lived, so if this is something that could affect the MCO post-install, then it is worth exploring further.

abays: The error described above was definitely something that appeared later in the deployment process.  I'll give these changes a go after this PR is merged and see if I encounter the same issue.  If I recall correctly, the MCO errors-out and thus the cluster operators never reach 100% availability.

ansible-ipi-install/roles/node-prep/tasks/70_enabled_fw_services.yml
```diff
@@ -1,19 +1,47 @@
...
+      become: yes
+      with_items:
+      - libvirt
+      - public
+  when: firewall != "iptables"
 
+- name: Configure iptables
+  block:
+    - name: Enable Firewall for HTTP
+      iptables:
```
e-minguez: AFAIK the iptables ansible module just modify the 'in memory' iptables rules, you need to store them to make them survive reboots.

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Create OpenShift Ignition Configs
+  shell: |
+    /usr/local/bin/openshift-baremetal-install --dir {{ dir }} create ignition-configs
+
+- name: Get Filetranspiler
+  get_url:
+    url: https://raw.githubusercontent.com/ashcrow/filetranspiler/master/filetranspile
```
sreichar: What will happen in a disconnected env?

rlopez133:
> Since this procedure is supposed to be automating teh best practices doc, can you help me understand why you could not use firewalld?
> 
> Also these changes should be reflected in the readme

@sreichar - The reason they don't use firewalld is specific to the lab environment that @dustinblack is using. Regarding documentation, I don't think we need to reflect this in the readme as the best practice should be firewalld. What we are trying to do here is just give flexibility. The default is firewalld (best practice) and iptables is only used if explicitly selected.

dustinblack:
> > Since this procedure is supposed to be automating teh best practices doc, can you help me understand why you could not use firewalld?
> > Also these changes should be reflected in the readme
> 
> @sreichar - The reason they don't use firewalld is specific to the lab environment that @dustinblack is using. Regarding documentation, I don't think we need to reflect this in the readme as the best practice should be firewalld. What we are trying to do here is just give flexibility. The default is firewalld (best practice) and iptables is only used if explicitly selected.

Correct -- For our internal shared lab environments, iptables is the norm so we need this flexibility. Roger and I worked together here to add the flexibility without and user overhead.

ansible-ipi-install/roles/node-prep/tasks/70_enabled_fw_services.yml
dustinblack: Good catch. I'll have a look at this.

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
dustinblack: This is a good point. I suppose it would need to be solved with documentation.

dustinblack:
> @dustinblack: PR needs rebase.

rebase complete

dustinblack:
> I've added a comment wrt iptables module.

fixed

ansible-ipi-install/roles/node-prep/tasks/70_enabled_fw_services.yml
```diff
@@ -1,19 +1,52 @@
...
+        chain: INPUT
+        ctstate: ESTABLISHED,RELATED
+        jump: ACCEPT
+      become: yes
+
+    - name: Save iptables configuration
+      shell: |
+        /usr/sbin/iptables-save > /etc/sysconfig/iptables
+      become: yes
```
e-minguez: I think you need to check if the iptables service is up otherwise enable it

dustinblack: That is handled in [ansible-ipi-install/roles/node-prep/tasks/60_enabled_services.yml](https://github.com/openshift-kni/baremetal-deploy/pull/139/files#diff-3465117d70f6807e3a48d826b378b960)

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
dustinblack: Chatting w/ @rlopez133 it sounds like disconnected needs to be its own feature effort.

ansible-ipi-install/roles/node-prep/tasks/70_enabled_fw_services.yml
```diff
@@ -1,19 +1,52 @@
 ---
-- name: Enable Firewall for HTTP
-  firewalld:
-    service: http
-    permanent: yes
-    state: enabled
-  become: yes
+- name: Configure firewalld
+  block:
+    - name: Enable Firewall for HTTP
```
rlopez133: I know this is nit but the task names look odd here when you run the play because you see the same task names multiple times. Here is an example when you don't have iptables but firewalld

```
TASK [node-prep : Enable Firewall for HTTP] ********************************************************************************
Tuesday 11 February 2020  20:01:20 -0600 (0:00:00.157)       0:03:00.174 ****** 
ok: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

TASK [node-prep : Open Firewall port 5000/tcp, zone libvirt and public] ****************************************************
Tuesday 11 February 2020  20:01:25 -0600 (0:00:04.766)       0:03:04.940 ****** 
ok: [worker-0.kni4.cloud.lab.eng.bos.redhat.com] => (item=libvirt)
ok: [worker-0.kni4.cloud.lab.eng.bos.redhat.com] => (item=public)

TASK [node-prep : Enable Firewall for HTTP] ********************************************************************************
Tuesday 11 February 2020  20:01:34 -0600 (0:00:08.763)       0:03:13.703 ****** 
skipping: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

TASK [node-prep : Open Firewall port 5000/tcp] *****************************************************************************
Tuesday 11 February 2020  20:01:34 -0600 (0:00:00.076)       0:03:13.780 ****** 
skipping: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

TASK [node-prep : Allow related and established connections] ***************************************************************
Tuesday 11 February 2020  20:01:34 -0600 (0:00:00.077)       0:03:13.858 ****** 
skipping: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

TASK [node-prep : Save iptables configuration] *****************************************************************************
Tuesday 11 February 2020  20:01:34 -0600 (0:00:00.076)       0:03:13.934 ****** 
skipping: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

TASK [node-prep : include_tasks] *******************************************************************************************
Tuesday 11 February 2020  20:01:34 -0600 (0:00:00.075)       0:03:14.009 ****** 
included: /home/rlopez/git_projects/baremetal-deploy/ansible-ipi-install/roles/node-prep/tasks/80_libvirt_pool.yml for worker-0.kni4.cloud.lab.eng.bos.redhat.com
```

Notice how you see "Enable Firewall for HTTP" and then 2 tasks later you see it again "Enable Firewall for HTP" but its skipped. I'd almost prefer it say for example "Enable {{ firewall }} for HTTP" -- this should translate to the appropriate value since the firewall variable is a fact.

ansible-ipi-install/roles/node-prep/vars/main.yml
```diff
@@ -1,7 +1,8 @@
 ---
 # vars file for node-prep
 package_list:
-  - firewalld
+  - "{{ firewall }}"
```
rlopez133: This one is interesting because right now firewall can only be firewalld or iptables. However, the logic in the 10_validation.yml is looking to see if iptables-service exists or not, if it does, we know you explicitly wanted to use iptables. In its current form it doesn't break anything because by default it uses firewalld and when iptables-service is installed, this variable changes to iptables but you likely already have iptables installed on the system since you had iptables-service. I just wanted to point that out since this won't install iptables-service which we only ever want if we actually going to use iptables.

ansible-ipi-install/roles/installer/tasks/25_customize_filesystem.yml
```diff
@@ -0,0 +1,55 @@
+---
+- name: Verify if {{ role_path }}/files/customize_filesystem/master/worker is empty
+  find:
+    paths: "{{ role_path }}/files/customize_filesystem/{{ item }}"
+  register: filesFound
+  with_items:
+    - "master"
+    - "worker"
+  delegate_to: localhost
```
dustinblack: @rlopez133 This change isn't working correctly. I need to root-cause and fix it.

rlopez133: @dustinblack what are you seeing?

dustinblack: The find needed to recurse in order to return the files nested in the directories, and the when logic for the block needed to be a little stronger. Fixed now.

ansible-ipi-install/roles/node-prep/tasks/40_bridge.yml
```diff
@@ -68,7 +68,15 @@
     mode: u=rw,g=r,o=r
   become: yes
 
-- name: Get DHCP for baremetal bridge
+- name: Set DNS priority for baremetal interface
   shell: |
-    /usr/sbin/ifdown {{ pub_nic }}; /usr/bin/pkill dhclient ; /usr/sbin/ifdown baremetal ; /usr/sbin/ifup {{ pub_nic }}; /usr/sbin/ifup baremetal
+    /usr/bin/nmcli con mod baremetal ipv4.dns-priority 10
```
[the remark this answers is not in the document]
dustinblack: I can probably remove this from here. It needs to be in our lab prerequisite instructions anyway.

Early on when you run the task to verify the DNS records, it will fail in our environment if the `/etc/resolv.conf` file has the nameserver entries from the public interface listed above our local dnsmasq server. So we need for our `baremetal` bridge to be configured with the higher (er, lower) `dns-priority` to ensure that our local nameserver gets listed first.

ansible-ipi-install/roles/node-prep/tasks/70_enabled_fw_services.yml
dustinblack: Fixed.

ansible-ipi-install/roles/node-prep/vars/main.yml
dustinblack: My argument for leaving this here is that if you have iptables in lieu of firewalld, you probably don't want the firewalld package to get installed. This will prevent that.

ansible-ipi-install/roles/node-prep/tasks/40_bridge.yml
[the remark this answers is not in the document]
rlopez133: Ok cool - yeah I vote to remove as its specific for your env

rlopez133:
cc @shardy

rlopez133:
cc @hardys

shardy:
Please remove me, @shardy, from your group. Wrong guy. 

_Steve

{via my mobile}

ansible-ipi-install/README.md
rlopez133: @abays - before I merge can you attempt this PR? From my view its merge ready but if we can see if you run into your issue before we merge makes my life easier :D 

```
git checkout -b dustinblack-shared_labs master
git pull https://github.com/dustinblack/baremetal-deploy.git shared_labs
```

abays: Sure, I will give it a try later today (I'm out at the moment).  My concern above was mainly with the example given as opposed to the general mechanism itself, though.  I otherwise agree that the PR looks good to go. :-)

abays: Well, good news: I'm not seeing the issue I was expecting.  Perhaps the behavior of the MCO has been changed.  Anyhow, it's working for me!

PR #146, opened 2020-02-06
Feature Enhancements of removing binaries from /usr/loca/bin and cleaning up tmp dirs

### LINKED ISSUE #144: Feature Enhancement of ansible-ipi-install - Clean up old tmp oc binary dir
Right now when the ansible-ipi-installer kicks off it creates a tmp directory that stores the oc binaries (oc, kubelet, openshift-baremetal-install). However, if an error occurred during deployment, this tmp file isn't cleaned up. Add logic that checks if these tmp dirs exist and remove them in order to save space on the local HD.

### LINKED ISSUE #145: Feature Enhancement of ansible-ipi-install - Remove binaries in /usr/local/bin on re-run of playbook
The playbook currently takes the oc binaries (oc, openshift-baremetal-install, kubelet) and places them in your /usr/local/bin directory. However, if say you wanted to reuse this provision host to install a new version of OCP, the existing playbook would notice that there is already an existing oc, openshift-baremetal-install,kubelet binary in /usr/local/bin and not place the newly created ones in /usr/local/bin. Solution is to create logic that checks if those files exist, and if they do, remove them prior to untarring the new binaries to be placed in the /usr/local/bin

rlopez133:
# Description
The playbook currently takes the oc binaries (oc, openshift-baremetal-install, kubelet) and places them in your /usr/local/bin directory. However, if say you wanted to reuse this provision host to install a new version of OCP, the existing playbook would notice that there is already an existing oc, openshift-baremetal-install,kubelet binary in /usr/local/bin and not place the newly created ones in /usr/local/bin. Solution is to create logic that checks if those files exist, and if they do, remove them prior to untarring the new binaries to be placed in the /usr/local/bin Fixes #145  

Right now when the ansible-ipi-installer kicks off it creates a tmp directory that stores the oc binaries (oc, kubelet, openshift-baremetal-install). However, if an error occurred during deployment, this tmp file isn't cleaned up. Add logic that checks if these tmp dirs exist and remove them in order to save space on the local HD. Fixes #144 

Fixes # (issue) #144 #145 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/10_get_oc.yml
+++ b/ansible-ipi-install/roles/installer/tasks/10_get_oc.yml
@@ -1,4 +1,30 @@
 ---
+- name: Find any old tmp dirs with OpenShift related binaries
+  find:
+    paths: /tmp
+    patterns: 'ansible.*'
+    file_type: directory
+  register: tmp_results
+
+- name: Delete any old tmp dirs with OpenShift related binaries
+  file:
+    path: "{{ item['path'] }}"
+    state: absent
+  loop: "{{ tmp_results['files'] }}"
+
+- name: Find any existing /usr/local/bin OpenShift binaries
+  find:
+    paths: /usr/local/bin
+    patterns: 'oc,openshift-baremetal-install,kubectl'
+  register: binary_results
+
+- name: Remove any existing /usr/local/bin OpenShift binaries
+  file:
+    path: "{{ item['path'] }}"
+    state: absent
+  loop: "{{ binary_results['files'] }}"
+  become: yes
+
 - name: Create tmp directory to store OpenShift binaries
   tempfile:
     state: directory
--- a/ansible-ipi-install/roles/installer/tasks/20_extract_installer.yml
+++ b/ansible-ipi-install/roles/installer/tasks/20_extract_installer.yml
@@ -24,3 +24,9 @@
     mode: '0755'
     remote_src: yes
   become: yes
+
+- name: Delete tmp dir as no longer needed
+  file:
+    path: "{{ tempdir }}"
+    state: absent
+


rlopez133:
> wouldnt it be better to use the binaries if there are around?

@karmab I thought about that, but I felt having it in /usr/local/bin so you can just use it without A) giving the full path or B) having to be in the dir where the binary exists is 1 more thing to do when its so easy to just put it in /usr/local/bin and done.

e-minguez:
I think it would be better to clean up the temporary directories after the successful download as well, so you don't need to wait to re run the installation to get rid of those temporary assets. So:
* First check if the temporary folders/files exist and if so remove them
* Download the binaries, etc.
* Clean the temporary folders/files

rlopez133:
> I think it would be better to clean up the temporary directories after the successful download as well, so you don't need to wait to re run the installation to get rid of those temporary assets. So:
> 
> * First check if the temporary folders/files exist and if so remove them
> * Download the binaries, etc.
> * Clean the temporary folders/files

* Left the check and delete of existing temp folders/files as is
* Download binaries section left as is
* Once the existing tempdir that is running the playbook is done copying the openshift-baremetal-install, go ahead and remove that directory as its no longer needed.

ansible-ipi-install/roles/installer/tasks/10_get_oc.yml
```diff
@@ -1,4 +1,30 @@
...
+    state: absent
+  loop: "{{ tmp_results['files'] }}"
+
+- name: Find any existing /usr/local/bin OpenShift binaries
+  find:
+    paths: /usr/local/bin
+    patterns: 'oc,openshift-baremetal-install,kubectl'
+  register: binary_results
+
+- name: Remove any existing /usr/local/bin OpenShift binaries
```
sreichar: since these binaries were copied (oc), should you also remove the original

rlopez133: @sreichar - I added a new file that will remove the entire /tmp directory once the openshift-baremetal-install binary has been copied.

Issue #153, opened 2020-02-11
Ansible: provisioning interface name needed in install-config


abays:
In 4.4, metal3-config.yaml is no longer required.  Instead, the designation of the provisioning interface name has been moved to install-config.yaml as `provisioningNetworkInterface`.  We should inject this value into the generated install-config.

PR #154, opened 2020-02-11
Inject prov_nic into install-config

abays:
# Description

4.4 requires `provisioningNetworkInterface` to be present in install-config.yaml to designate the name of the provisioning interface used during deployment.  Failure to do so seems to result in a default of `ens3` being used, which won't work for every baremetal environment.  We can take `prov_nic` and inject it into the install-config j2 template to fix this problem.

Fixes https://github.com/openshift-kni/baremetal-deploy/issues/153

## Type of change

## Testing

I run an ansible-ipi-install against my baremetal environment without injecting the provisioning interface name in the install-config.yaml, and I see that the deployment incorrectly defaults to `ens3`.  I then run the ansible-ipi-install again, this time with the install-config j2 template changes for injecting the provisoning nic name, and the proper interface appears within deployment objects.

**Test Configuration**:

- Versions:
- Hardware: OCTO NFVHA Supermicro lab

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/templates/install-config.j2
+++ b/ansible-ipi-install/roles/node-prep/templates/install-config.j2
@@ -17,7 +17,9 @@ platform:
     apiVIP: {{ apivip }}
     ingressVIP: {{ ingressvip }}
     dnsVIP: {{ dnsvip }}
+{% if (release_version[0]|int >= 4) and (release_version[2]|int > 3) %}
     provisioningNetworkInterface: {{ prov_nic }}
+{% endif %}
     hosts:
 {% for host in groups['masters'] %}
       - name: {{ hostvars[host]['name'] }}
--- a/ansible-ipi-install/roles/node-prep/templates/install-config.j2
+++ b/ansible-ipi-install/roles/node-prep/templates/install-config.j2
@@ -17,6 +17,7 @@ platform:
     apiVIP: {{ apivip }}
     ingressVIP: {{ ingressvip }}
     dnsVIP: {{ dnsvip }}
+    provisioningNetworkInterface: {{ prov_nic }}
     hosts:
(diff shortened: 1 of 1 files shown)


abays:
> To my knowledge this is only related to 4.4 correct? Also I think I want to add more logic here regarding version i.e. if version X used include this, otherwise dont include.

Yes, that is true.  We could certainly add some versioning conditionals.  In my testing, however, it didn't seem to matter than `provisioningNetworkInterface` was injected into a 4.3 install-config.yaml.  It seemed to just be ignored.

rlopez133:
> > To my knowledge this is only related to 4.4 correct? Also I think I want to add more logic here regarding version i.e. if version X used include this, otherwise dont include.
> 
> Yes, that is true. We could certainly add some versioning conditionals. In my testing, however, it didn't seem to matter than `provisioningNetworkInterface` was injected into a 4.3 install-config.yaml. It seemed to just be ignored.

So talking to Stephen Benjamin, he mentioned right now if the installer doesn't recognize a key/value pair in the install-config.yml, it ignores it but he mentioned this may not be the case always.

abays:
> > > To my knowledge this is only related to 4.4 correct? Also I think I want to add more logic here regarding version i.e. if version X used include this, otherwise dont include.
> > 
> > 
> > Yes, that is true. We could certainly add some versioning conditionals. In my testing, however, it didn't seem to matter than `provisioningNetworkInterface` was injected into a 4.3 install-config.yaml. It seemed to just be ignored.
> 
> So talking to Stephen Benjamin, he mentioned right now if the installer doesn't recognize a key/value pair in the install-config.yml, it ignores it but he mentioned this may not be the case always.

Okay, sounds like a version check is a good idea here, then. :+1:

rlopez133:
fixes #160

PR #161, opened 2020-02-12
Do not generate metal3-config for 4.4+

abays:
# Description

The Ansible logic should not generate a metal3-config.yaml file for 4.4+ deployments, as this ConfigMap is no longer required.

Fixes https://github.com/openshift-kni/baremetal-deploy/issues/151

## Type of change

## Testing

1. Set version variable to 4.3.XXXXXX and run the playbook.  Verify that metal3-config.yaml is created in ~/clusterconfigs.
2. Remove any existing metal3-config.yaml from ~/clusterconfigs.  Set version variable to 4.4.XXXXXX and run the playbook.  Verify that that metal3-config.yaml is NOT created in ~/clusterconfigs.

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
+++ b/ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
@@ -23,3 +23,4 @@
     group: "{{ ansible_user }}"
     mode: '0644'
     remote_src: yes
+  when: (release_version[0]|int >= 4) and (release_version[2]|int <= 3)
--- a/ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
+++ b/ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
@@ -23,4 +23,4 @@
     group: "{{ ansible_user }}"
     mode: '0644'
     remote_src: yes
-  when: (release_version[0]|int >= 4) and (release_version[2]|int <= 3)
+  when: (release_version[0]|int == 4) and (release_version[2]|int <= 3)
--- a/ansible-ipi-install/roles/installer/tasks/main.yml
+++ b/ansible-ipi-install/roles/installer/tasks/main.yml
@@ -3,7 +3,7 @@
 - include_tasks: 10_get_oc.yml 
 - include_tasks: 20_extract_installer.yml
 - include_tasks: 30_create_metal3.yml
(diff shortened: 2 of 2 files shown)


ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
```diff
@@ -23,3 +23,4 @@
     group: "{{ ansible_user }}"
     mode: '0644'
     remote_src: yes
+  when: (release_version[0]|int >= 4) and (release_version[2]|int <= 3)
```
rlopez133: shouldn't this be `release_version[0]|int >= 4) and (release_version[2]|int > 3` ? Either way theres logic here that for example in your case if you are version 5.4 it would break or in the logic I provided 5.2 it would break.

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -3,6 +3,7 @@
 - include_tasks: 10_get_oc.yml 
 - include_tasks: 20_extract_installer.yml
 - include_tasks: 30_create_metal3.yml
+  when: (release_version[0]|int >= 4) and (release_version[2]|int <= 3)
```
rlopez133: same as above comment.

abays: Yeah, I was just thinking about the conditional while grabbing lunch.  Needs a tweak. Will fix!

rlopez133: Just FYI, same can be said about the #153 where the same logic was applied for install-config.j2 which we merged in already.

abays: Should I just make that change on this PR as well?

rlopez133: no, after we merge this. lets just make a separate PR for the #153

abays:
Do we have to consider OpenShift 3 in these conditionals?  I would think that major version would never be used.

rlopez133:
> Do we have to consider OpenShift 3 in these conditionals? I would think that major version would never be used.

I think its safe to say this is 4.x . Honestly 4.3+ would be good enough IMO

Issue #167, opened 2020-02-18
Ansible: hardware_profile not found in host vars


abays:
If you clone a fresh copy of baremetal-deploy and happen to leave `hardware_profile` out of your inventory hosts definitions, you will hit this error:

```
TASK [node-prep : Generate install-config.yaml] *******************************************************************************************************************************************************************
Tuesday 18 February 2020  07:37:28 -0500 (0:00:00.198)       0:01:38.178 ******                                                                                                                                     
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'hardware_profile'"}
```

Since this is the first time running the playbook, there is no `hardware_profile` attribute saved in the Ansible cache for the hosts -- thus you encounter this error if you did not explicitly set the variable.  We should first check for `hardware_profile`'s presence in the host vars before trying to reference it.

PR #168, opened 2020-02-18
Check for hardware_profile in host vars

abays:
# Description

If the user fails to include `hardware_profile` in their inventory hosts in the Ansible tool, and they happen to be using a fresh clone of the repo, then Ansible will error-out when trying to access the non-existent `hardware_profile` in the host vars during install-config.yaml generation (as the playbook has never been executed before, the cached host vars lack the `hardware_profile` attribute).  We should check that the attribute is defined before trying to reference it.

Fixes https://github.com/openshift-kni/baremetal-deploy/issues/167

## Type of change

## Testing

- Clone the repo and create your hosts file, leaving `hardware_profile` out
- Run the playbook, and ensure the the "Generate install config" task executes without error

**Test Configuration**:

- Versions:
- Hardware:

## Checklist


e-minguez:
I forgot to test this without the variable, thanks for the fix @abays

abays:
> I forgot to test this without the variable, thanks for the fix @abays

@e-minguez No problem!  Glad to help.

PR #169, opened 2020-02-20
Ansible: RHCOS image caching

### LINKED ISSUE #176: [RFE] RHCOS caching
We want to support pre-caching RHCOS images for 4.3/4.4 via our Ansible tool. This includes bootstrapOSImage and clusterOSImage

abays:
# Description

We want to support pre-caching RHCOS images for 4.3/4.4 via our Ansible tool, such that the bootstrap VM can download these images from the local store rather than having to go out to the Internet.  These changes to the Ansible tool allow this.

Fixes #176 

## Type of change

## Testing

1. Set version to `latest-4.3` and use_rhcos_image_cache to `false`.  Execute playbook and ensure that the bootstrap downloads the RHCOS images from the Internet.
2. Set version to `latest-4.3` and use_rhcos_image_cache to `true`.  Execute playbook and ensure that the bootstrap downloads the RHCOS images from the local Apache webserver on the provisioning host.
3. Set version to `latest-4.4` and use_rhcos_image_cache to `false`.  Execute playbook and ensure that the bootstrap downloads the RHCOS images from the Internet.
4. Set version to `latest-4.4` and use_rhcos_image_cache to `true`.  Execute playbook and ensure that the bootstrap downloads the RHCOS images from the local Apache webserver on the provisioning host.

**Test Configuration**:

- Versions:
- Hardware:

## Checklist


ansible-ipi-install/README.md
```diff
@@ -374,6 +377,17 @@ included when the OpenShift manifests are created.
...
 
+## Pre-caching RHCOS Images
+By default, the playbook will pre-download RHCOS images prior to actual cluster
+deployment.  It places these images in an Apache web server on the provisioning
+host, and modifies `metal3-config.yaml` or `install-config.yaml` (depending on 
+whether the version is set to 4.3 or 4.4+) to instruct the bootstrap to
+download the images from that web server during deployment.  
+
+If you wish to force the bootstrap to download the RHCOS images from the Internet,
+set the `use_rhcos_image_cache` variable to false (in your hosts file, via a
```
rlopez133: Instead of forcing user to set value of `use_rhcos_image_cache` to false, lets set the variable inside node-prep/defaults/main.yml to `use_rhcos_image_cache: false` . This way in the inventory/hosts.sample file we can just have a commented line that shows #use_rhcos_image_cache: true with a brief description. 

Example:
#(Optional) Enable playbook to pre-download RHCOS images prior to cluster deployment. Default is false.
#use_rhcos_image_cache=true

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 # defaults file for installer
+use_rhcos_image_cache: true
```
rlopez133: its complaining about no newline at end of file here.

rlopez133: Also, I'd think we wouldn't want this set to true by default?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+
+- name: Download {{ rhcos_qemu_uri }} for cache (4.4+ only)
+  get_url:
+    url:  "{{ rhcos_path }}/{{ rhcos_qemu_uri }}"
+    dest: "/var/www/html/{{ rhcos_qemu_uri }}"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    mode: '0755'
+  when: ((release_version[0]|int > 4) or (release_version[2]|int > 3)) and (rhcos_qemu_stat.stat.exists == false or rhcos_qemu_stat.stat.checksum != rhcos_qemu_sha256)
+  become: yes
```
rlopez133: because you use `ansible_user` here and become: yes, this will make the owner and group root/root (FYI). Is this OK?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+
+- name: Download {{ rhcos_openstack_uri }} for cache
+  get_url:
+    url:  "{{ rhcos_path }}/{{ rhcos_openstack_uri }}"
+    dest: "/var/www/html/{{ rhcos_openstack_uri }}"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    mode: '0755'
+  when: (rhcos_openstack_stat.stat.exists == false or rhcos_openstack_stat.stat.checksum != rhcos_openstack_sha256)
+  become: yes
```
rlopez133: because you use ansible_user here and become: yes, this will make the owner and group root/root (FYI). Is this OK?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+        path: "/var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}"
+        state: directory
+        owner: "{{ ansible_user }}"
+        group: "{{ ansible_user }}"
+        mode: '0755'
+      become: yes
+
+    # unarchive module does NOT support non-tar gz files, so we have to use shell here
+    - name: Extract {{ rhcos_openstack_uri }} into image cache sub-dir
+      shell: "gunzip -ck /var/www/html/{{ rhcos_openstack_uri }} > /var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}/{{ rhcos_openstack_uri_minus_gz }}"
```
rlopez133: do we need to put abs path here for gunzip?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+      become: true
+
+    - name: Convert image into compressed qcow2
+      shell: "qemu-img convert -O qcow2 -c /var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}/{{ rhcos_openstack_uri_minus_gz }} /var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}/{{ rhcos_compressed_qcow2 }}"
```
rlopez133: do we need to put abs path for qemu-img?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+      become: yes
+    
+    - name: Get md5sum of new compressed qcow2 image
+      shell: "md5sum /var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}/{{ rhcos_compressed_qcow2 }} | awk '{print $1}'"
```
rlopez133: do we need to put abs path for md5sum?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+        dest: "/var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}/{{ rhcos_compressed_qcow2 }}.md5sum"
+        owner: "{{ ansible_user }}"
+        group: "{{ ansible_user }}"
+        mode: '0644'
+      become: yes
+  when: (release_version[0]|int == 4) and (release_version[2]|int <= 3) and ((rhcos_openstack_stat.stat.exists == false or rhcos_openstack_stat.stat.checksum != rhcos_openstack_sha256))
+
+- name: Override existing rhcos_path variable
+  set_fact:
+    rhcos_path: "http://172.22.0.1/"
```
rlopez133: Does this 24_* script always run? For my FYI purposes why does it need to change from `rhcos_path: "{{ rhcos_json.json | json_query('baseURI') }}"` ?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+  with_items:
+    - "{{ dir }}/install-config.yaml"
+    - "{{ dir }}/install-config.yaml.bkup"
+  when: (release_version[0]|int > 4) or (release_version[2]|int > 3)
+
+- name: Make sure httpd is started
+  systemd:
+    state: started
+    name: httpd
+  become: yes
```
rlopez133: no newline at end of file.

ansible-ipi-install/roles/installer/templates/metal3-config.j2
```diff
@@ -13,5 +13,5 @@ data:
   ironic_inspector_endpoint: http://{{ prov_ip }}:5050/v1/
   provisioning_interface: {{ prov_nic }}
   provisioning_ip: {{ prov_ip }}/24
-  rhcos_image_url: {{ rhcos_path }}{{ rhcos_uri }}
+  rhcos_image_url: {{ rhcos_path }}{{ rhcos_openstack_uri }}
```
rlopez133: do we want to have the word openstack in here?

ansible-ipi-install/roles/installer/templates/rhcos-image-md5sum.j2
```diff
@@ -0,0 +1 @@
+{{ rhcos_image_md5sum.stdout }}
```
rlopez133: newline

ansible-ipi-install/roles/installer/defaults/main.yml
abays: Ack

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: It does do as you note, but it doesn't cause a problem.  I wanted to save a couple plays for setting the permissions, but we can add them if you think it would be wise.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: If we don't, then it will look for the file in the Ansible user's home directory (which is not where it is located).  We're using a bare-bones `httpd` installation that uses `/var/www/html/` as the content directory, which is why it is hardcoded here.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: Same concern here as above.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: 24_* runs when `use_rhcos_image_cache` is true.  `rhcos_path` points to the base URL where the images are located.  Initially, this value is some external URL.  It's changed here to the provisioning bridge IP for use when setting the actual cache URLs later.  An alternative would be to set a new fact that serves as the images' base URL, to differentiate it from the URL used to download the images from the Internet.

ansible-ipi-install/roles/installer/templates/metal3-config.j2
abays: The images downloaded are in fact the `openstack` ones, i.e. `rhcos-44.81.202001241431.0-openstack.x86_64.qcow2.gz`.  Would could call it something else though.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
rlopez133: Since in the future they will give the ability to change the IP from 172.22.0.1 (tmk), I think what we should do is likely capture the IP from the provisioning nic and then have it set that value. Where this will be tricky is at some point, we will need to differentiate between ipv4 and ipv6 and i figured i'd mention it here in case you had a bright idea for that. However, for now,  ipv4 is fine.

ansible-ipi-install/roles/installer/templates/metal3-config.j2
rlopez133: id almost prefer to call it rhcos_uri and not have anything relating to osp. For all we know they will change the image to something else but we wouldn't need to worry since ours is rhcos_uri versus calling out osp in the name.

abays: Understood.  Another thing I forgot to mention was that I used `openstack` to differentiate from the bootstrap image, which is `qemu`.  How about we keep `rhcos_qemu_uri` and leave the other one as it was before this PR: `rhcos_uri`?

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: > Andrew,
> 
> I think breaking this into 2 part is called for. One part that does the downloading, another that uses a cache.
> 

@sreichar Could you clarify further regarding breaking it into two parts?  Do you mean perhaps having a separate playbook/role that downloads the images?  It's true that right now a single boolean flag determines whether the images are downloaded and whether the metal3-config/install-config is set to use the pre-downloaded images.  We could easily break that up into separate flags (or sets of download and cache parameters).

> We are hoping to have a lab shared cache and can see others doing the same. So this requires checking is a caching mechanism is running and if it is not, then starting it.
> 

I think I get what you're saying here.  Perhaps another group type in the hosts called "cache" that could contain the IP of the machine that would host the httpd cache, and then a separate role to go with it.

> The using the cache will also need to the option to specify the caching source. This also means having a check that is image is available before using the cache. Obviously having default to the provision host would be the best.

Understood.  One thing to note, though, regarding checking if the image is available in the cache before using it: the bootstrap will fail-over to the standard image source (the Internet) if the supplied image source (some cache URL via metal3-config or install-config) isn't available.  In other words, the deployment already does this check for us.

ansible-ipi-install/README.md
rlopez133: Can we add this also to the inventory/hosts.sample file?

abays: Ack.  I think it's there already.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
rlopez133: Chiming in here:

1) @abays @sreichar regarding breaking into parts, I see three flags. One flag for downloading image, 1 flag for caching, 1 flag for caching host. All flags being false by default. However, I'd see the logic as:
 IF downloading the image is set to true, then caching will be set to true. If both are true, see if a caching host has been specified, if yes, use that location otherwise create a container on the provision host to host the images. 

IF downloading the image is false and caching is set to true - look to see if there is a caching host specified. Error if no host specified and ask to either specify one or enable downloading images. 

IF caching is set to false, then downloading and cache host var auto get set to false via a set_fact and no caching is needed.

rlopez133: To add:
We will need an additional group in the inventory/hosts file such as [cache-host] . We will check the length of this group in the validation.yml, if the length is 0 , we know to use the provisioning host for caching. If length > 0 then we know to use what is specified as the cache host. If length is > 1, user entered too many and ask them to only specify one cache host.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,125 @@
...
+  register: rhcos_qemu_stat
+  when: (release_version[0]|int > 4) or (release_version[2]|int > 3)
+
+- name: Download {{ rhcos_qemu_uri }} for cache (4.4+ only)
+  get_url:
+    url:  "{{ rhcos_path }}/{{ rhcos_qemu_uri }}"
+    dest: "/var/www/html/{{ rhcos_qemu_uri }}"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    mode: '0755'
```
rlopez133: add timeout: 3600

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,125 @@
...
+    get_checksum: yes
+  register: rhcos_openstack_stat
+
+- name: Download {{ rhcos_openstack_uri }} for cache
+  get_url:
+    url:  "{{ rhcos_path }}/{{ rhcos_openstack_uri }}"
+    dest: "/var/www/html/{{ rhcos_openstack_uri }}"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    mode: '0755'
```
rlopez133: add timeout: 3600

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: I think @rlopez133 's suggestions are a great starting part.  Let's continue to flesh this out.

ansible-ipi-install/README.md
```diff
@@ -374,6 +382,20 @@ included when the OpenShift manifests are created.
 
 NOTE: By default this directory is empty. 
 
+## Pre-caching RHCOS Images
+If you wish to set up a local cache of RHCOS images on your provisioning host,
+set the `cache_enabled` variable to `True` in your hosts file (make sure you use
+`True` and not `true`).  When requested, the playbook will pre-download RHCOS
```
rlopez133: This doesn't matter because I take care of this via logic.

ansible-ipi-install/README.md
```diff
@@ -374,6 +382,20 @@ included when the OpenShift manifests are created.
 
 NOTE: By default this directory is empty. 
 
+## Pre-caching RHCOS Images
+If you wish to set up a local cache of RHCOS images on your provisioning host,
+set the `cache_enabled` variable to `True` in your hosts file (make sure you use
+`True` and not `true`).  When requested, the playbook will pre-download RHCOS 
+images prior to actual cluster deployment.  It places these images in an Apache 
+web server container on the provisioning host and modifies `install-config.yaml` 
+to instruct the bootstrap to download the images from that web server during
```
rlopez133: "to instruct the bootstrap VM"

ansible-ipi-install/README.md
```diff
@@ -374,6 +382,20 @@ included when the OpenShift manifests are created.
...
+images prior to actual cluster deployment.  It places these images in an Apache 
+web server container on the provisioning host and modifies `install-config.yaml` 
+to instruct the bootstrap to download the images from that web server during 
+deployment.  
+
+NOTE: If you set the `clusterosimage` and `bootstraposimage` variables, then
+`cache_enabled` will automatically be set to `False`, since the combined  
+presence of these values indicates that your RHCOS images are already available
+elsewhere.
+
```
rlopez133: Would change the sentence to: 
"... be set to `False`. Setting these variables leaves the responsibility to the end user in ensuring the RHCOS images are readily available and accessible to the provision host."

ansible-ipi-install/roles/node-prep/vars/main.yml
```diff
@@ -16,3 +16,5 @@ package_list:
   - policycoreutils-python3
   - https://trunk.rdoproject.org/rhel8-master/deps/latest/Packages/python3-crypto-2.6.1-18.el8ost.x86_64.rpm
   - https://trunk.rdoproject.org/rhel8-master/deps/latest/Packages/python3-pyghmi-1.0.22-2.el8ost.noarch.rpm
+  - httpd
+  - gzip
```
rlopez133: because we doing the container, do we need these packages anymore? (httpd and gzip)

abays: Ah, good catch.  Nope!

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,87 @@
...
+    name: node-prep
+    tasks_from: 100_create-install-config
+
+- name: Start RHCOS image cache container
+  podman_container:
+    name: rhcos_image_cache
+    image: httpd
+    state: started
+    exposed_ports:
+      - 80
```
e-minguez: I'm not sure if opening the firewall is required for this to work.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,87 @@
...
+# URIs.
+- name: Re-do install-config.yaml generation with cache values
+  include_role:
+    name: node-prep
+    tasks_from: 100_create-install-config
+
+- name: Start RHCOS image cache container
+  podman_container:
+    name: rhcos_image_cache
+    image: httpd
```
e-minguez: I think we better use the RHEL8 httpd image instead the docker.io one. See https://access.redhat.com/containers/#/registry.access.redhat.com/rhel8/httpd-24

e-minguez: And if so, the volumes shall be `"{{ provision_cache_store }}:/var/www"`

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,87 @@
...
+    state: started
+    exposed_ports:
+      - 80
+    volumes:
+      - "{{ provision_cache_store }}:/usr/local/apache2/htdocs"
+    network: host
```
e-minguez: Does this mean `--net=host` ?

abays: Yes, it is equivalent to that.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,124 @@
...
+    path: "{{ item }}"
+    insertafter: "dnsVIP:"
+    block: |4
+            bootstrapOSImage: "{{ rhcos_path }}{{ rhcos_qemu_uri }}?sha256={{ rhcos_qemu_sha256 }}"
+            clusterOSImage: "{{ rhcos_path }}{{ rhcos_openstack_uri }}?sha256={{ rhcos_openstack_sha256 }}"
+  with_items:
+    - "{{ dir }}/install-config.yaml"
+    - "{{ dir }}/install-config.yaml.bkup"
+  when: (release_version[0]|int > 4) or (release_version[2]|int > 3)
+
```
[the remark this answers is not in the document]
abays: Good call, but we do actually already open 80 in 70_enabled_fw_services.yml, whether the user is using firewalld or iptables.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: Ack.  Let's use the RHEL 8 image as you suggest.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,87 @@
...
+    dest: "{{ provision_cache_store }}{{ rhcos_uri }}"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    mode: '0644'
+    timeout: 3600
+  when: (clusterosimage is not defined or clusterosimage|length < 1) and (rhcos_stat.stat.exists == false or rhcos_stat.stat.checksum != rhcos_sha256)
+
+- name: Set bootstrap image URL override if not provided by the user
+  set_fact:
+    bootstraposimage: "http://172.22.0.1/{{ rhcos_qemu_uri }}?sha256={{ rhcos_qemu_sha256 }}"
```
rlopez133: Lets not use hardcoded 172.22.0.1 as this may change in the future instead lets use 
`{{ ansible_provisioning.ipv4.address }}`

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -0,0 +1,87 @@
...
+  when: bootstraposimage is not defined or bootstraposimage|length < 1
+
+- name: Set cluster image URL override if not provided by the user
+  set_fact:
+    clusterosimage: "http://172.22.0.1/{{ rhcos_uri }}?sha256={{ rhcos_sha256 }}"
```
rlopez133: lets not use hardcoded 172.22.0.1, instead lets use
`{{ ansible_provisioning.ipv4.address }}`

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -2,6 +2,10 @@
 # tasks file for installer
 - include_tasks: 10_get_oc.yml 
 - include_tasks: 20_extract_installer.yml
+- include_tasks: 23_rhcos_image_paths.yml
+  when: ((release_version[0]|int == 4) and (release_version[2]|int <= 3)) or cache_enabled == true
```
rlopez133: just confirming here the behavior. So we expect 23_* yml to run when version is 4.3 or lower or cache is true

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -2,6 +2,10 @@
...
+- include_tasks: 24_rhcos_image_cache.yml
+  when: cache_enabled == true
```
rlopez133: just confirming behavior: we expect 24_* to always run if cached_enabled is true? Btw I'd change these to uppercase True

ansible-ipi-install/roles/installer/templates/rhcos-image-md5sum.j2
rlopez133: where does this get used?

ansible-ipi-install/roles/installer/tasks/main.yml
abays: Yes, 23_* runs if the cache is enabled or if the version is 4.3.  4.3 needs 23_* regardless of cache due to the RHCOS image paths needed in metal3-config.yaml.

ansible-ipi-install/roles/installer/tasks/main.yml
abays: That's correct.

ansible-ipi-install/roles/installer/templates/rhcos-image-md5sum.j2
abays: Good catch.  We no longer need this template.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
abays: @e-minguez Unfortunately the RHEL 8 httpd image in the redhat.io seems to require authentication.  That would add a barrier to entry to using the cache that we probably don't want.  So for now it looks like we'll stick with the Docker version.

e-minguez: Bummer. What about using the CentOS one? https://registry.centos.org/centos/httpd-24-centos7/latest or https://registry.centos.org/centos/httpd/latest

abays: Sure, let me give that a try

abays: @e-minguez It worked!  Thanks for the suggestion :+1:

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -30,6 +30,10 @@ build=""
 # Provisioning IP address (default value)
 prov_ip=172.22.0.3
 
+#(Optional) Enable playbook to pre-download RHCOS images prior to cluster deployment and use them as a local
+cache.  Default is false.
```
rlopez133: need a '#' in front to make it a comment

ansible-ipi-install/roles/installer/templates/metal3-config.j2
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: openshift-machine-api
 data:
 {% if clusterosimage is defined and clusterosimage|length %}
-  cache_url: {{ (clusterosimage.split("/")[:-1])|join("/") }}
+  cache_url: ''
 {% else %}
   cache_url: {{ rhcos_path }}
```
rlopez133: actually i think we dont need this if statement at all and just make cache_url: ''

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
maxamillion: This could alternatively be done like this:
```yaml
- name: Get md5sum of new compressed qcow2 image
  stat:
    path: "/var/www/html/images/{{ rhcos_openstack_uri_minus_gz }}/{{ rhcos_compressed_qcow2 }}"
    checksum_algorithm: md5
  register: compressed_qcow_stat
```

From there the `checksum` attribute of the `compressed_qcow_stat` variable could be referenced as `compressed_qcow_stat['checksum']`. This is mostly a stylistic preference over having shell one-liners and I don't know that one is inherently more or less correct than the other but I figured I'd point it out as an option.

rlopez133:
closing PR as we merged it in via https://github.com/openshift-kni/baremetal-deploy/pull/194

PR #171, opened 2020-02-21
Enable async install and looping of wait-for commands

dustinblack:
# Description

This change to the ansible deploy_ocp play allows for async'ing the initial installer to the background and then separately running wait-for commands for both bootstrap-complete and install-complete. This has two positive effects:

1. This extends (and allows for further extension of) the installer timeout, which is currently not adjustable at the installer itself and can be too short for larger bare metal deployments.
2. This provides more incremental feedback to the user regarding the status of the installation process.

## Type of change

## Testing

- [x] Tested an end-to-end IPI installation with the ansible automation, confirming that the async installer still allows the foreground wait-for commands and that the install completes successfully.

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
+++ b/ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
@@ -8,4 +8,41 @@
 - name: Deploy OpenShift Cluster
   shell: |
     /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug create cluster
+  when:
+    - increase_bootstrap_timeout is not defined
+    - increase_install_timeout is not defined
+  tags: install
+
+- name: Deploy OpenShift Cluster with extended timeouts
+  block:
+    - name: Run OpenShift Cluster install as async task
+      shell: |
+        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug create cluster
+      async: 3600
+      poll: 0
+      ignore_errors: yes
+      register: installer_result
+
+    - name: Wait for kubeconfig file
+      wait_for:
+        path: "{{ dir }}/auth/kubeconfig"
+        timeout: 90
+        msg: Timeout waiting for kubeconfig file
+
+    - name: Wait for Bootstrap Complete
+      shell: |
+        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug wait-for bootstrap-complete
+      register: wait_for_bootstrap_result
+      until: wait_for_bootstrap_result is succeeded
+      retries: "{{ increase_bootstrap_timeout|int }}"
+      delay: 1
+
+    - name: Wait for Install Complete
+      shell: |
+        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug wait-for install-complete
+      register: wait_for_install_result
+      until: wait_for_install_result is succeeded
+      retries: "{{ increase_install_timeout|int }}"
+      delay: 1
+  when: increase_bootstrap_timeout is defined or increase_install_timeout is defined
   tags: install
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -267,6 +267,29 @@ prov_ip=172.22.0.3
 # (Optional) A list of clock servers to be used in chrony by the masters and workers
 #clock_servers=["pool.ntp.org","clock.redhat.com"]
 
+# (Optional) Provide HTTP proxy settings
+#http_proxy=http://USERNAME:PASSWORD@proxy.example.com:8080
+
+# (Optional) Provide HTTPS proxy settings
+#https_proxy=https://USERNAME:PASSWORD@proxy.example.com:8080
+
+# (Optional) comma-separated list of hosts, IP Addresses, or IP ranges in CIDR format
+# excluded from proxying
+# NOTE: OpenShift does not accept '*' as a wildcard attached to a domain suffix
+# i.e. *.example.com
+# Use '.' as the wildcard for a domain suffix as shown in the example below.
+# i.e. .example.com
+#no_proxy_list="172.22.0.0/24,.example.com"
+
+# The default installer timeouts for the bootstrap and install processes may be too short for some baremetal
+# deployments. The variables below can be used to extend those timeouts.
+
+# (Optional) Increase bootstrap process timeout by N iterations.
+#increase_bootstrap_timeout=2
+
+# (Optional) Increase install process timeout by N iterations.
+#increase_install_timeout=2
+
 ######################################
 # Vars regarding install-config.yaml #
 ######################################
--- a/ansible-ipi-install/inventory/hosts.sample
+++ b/ansible-ipi-install/inventory/hosts.sample
@@ -64,6 +64,14 @@ prov_ip=172.22.0.3
 # i.e. .example.com
 #no_proxy_list="172.22.0.0/24,.example.com"
 
+# The default installer timeouts for the bootstrap and install processes may be too short for some baremetal
+# deployments. The variables below can be used to extend those timeouts.
+
+# (Optional) Increase bootstrap process timeout by N iterations.
+#increase_bootstrap_timeout=2
+
+# (Optional) Increase install process timeout by N iterations.
+#increase_install_timeout=2
 
 ######################################
 # Vars regarding install-config.yaml #


e-minguez:
I like this very much, but I would like to have this as optional as I believe most folks would like to run the installation in a single step. Also, the timeouts have been increased in 4.4 and 4.3 lately and I had a couple of successful runs with masters + workers installed in a single step.
My 2 cents

rlopez133:
> I like this very much, but I would like to have this as optional as I believe most folks would like to run the installation in a single step. Also, the timeouts have been increased in 4.4 and 4.3 lately and I had a couple of successful runs with masters + workers installed in a single step.
> My 2 cents

@dustinblack - considering that the installer team is already increasing the timeout values and based on what @e-minguez mentioned to me regarding that customer/user ain't going to know about running the timeout, I think we should leave it out.

rlopez133:
> > I like this very much, but I would like to have this as optional as I believe most folks would like to run the installation in a single step. Also, the timeouts have been increased in 4.4 and 4.3 lately and I had a couple of successful runs with masters + workers installed in a single step.
> > My 2 cents
> 
> @dustinblack - considering that the installer team is already increasing the timeout values and based on what @e-minguez mentioned to me regarding that customer/user ain't going to know about running the timeout, I think we should leave it out.

@e-minguez @dustinblack -- thinking about this further. Maybe not including it is not the answer considering if it is an option with the current deployer. Maybe what we should do instead is what @e-minguez mentioned of making it optional. We can create two flags in the inventory/hosts.sample file labeled `increase_bootstrap_timeout` and `increase_installer_timeout`

We can have wording as such in the README and the inventory/hosts.sample:
```
# (Optional) Add an additional 30 minutes for the bootstrap process to complete. Default False.
#increase_bootstrap_timeout=True
#(Optional) Add an additional 30 minutes for the installer process to complete. Default False.
#increase_installer_timeout=True
```

@dustinblack within the 10_validation script, create the following set_fact in case someone uses the value 'true' versus 'True' because within the inventory/hosts file if the value is 'true' it will use the string value versus the boolean value. 
Here is a snippet of the code you need for 10_validation.yml

```
- name: Conditions to set increase_bootstrap_timeout to true
  set_fact:
    increase_bootstrap_timeout: true
  when: (increase_bootstrap_timeout == 'true')

- name: Conditions to set increase_installer_timeout to true
  set_fact:
   increase_installer_timeout: true
  when: (increase_installer_timeout == 'true')

```

dustinblack:
> @dustinblack within the 10_validation script, create the following set_fact in case someone uses the value 'true' versus 'True' because within the inventory/hosts file if the value is 'true' it will use the string value versus the boolean value.
> Here is a snippet of the code you need for 10_validation.yml
> 
> ```
> - name: Conditions to set increase_bootstrap_timeout to true
>   set_fact:
>     increase_bootstrap_timeout: true
>   when: (increase_bootstrap_timeout == 'true')
> 
> - name: Conditions to set increase_installer_timeout to true
>   set_fact:
>    increase_installer_timeout: true
>   when: (increase_installer_timeout == 'true')
> ```

My preference would be to put this the `vars/main.yml` file for the role and use the `default` filter.

```
increase_bootstrap_timeout: {{ increase_bootstrap_timeout | default('false') }}
increase_installer_timeout: {{ increase_installer_timeout | default('false') }}
```

We may also want to consider making this a time value instead of a boolean.

dustinblack:
Have a look at the changes @rlopez133 @e-minguez

dustinblack:
> I'd say to have this as optional as I believe most folks would like to run the installation in a single step, so no async and retries if not needed (some conditional that if the `increase_bootstrap_timeout` and `increase_install_timeout` variables are 0, then run the installation in sync mode, otherwise have a block to run the async mode).
> My 2 cents.

I disagree. Part of the benefit of running the main installer async and then the wait-for's separately is that the user gets at least a little more incremental feedback from Ansible. Currently, the installer task can take up to an hour (maybe longer, depending on release version) with zero feedback on the status. By reporting the bootstrap-complete state we get intermediate feedback, even if we don't do an re-runs of the tasks.

e-minguez:
> > I'd say to have this as optional as I believe most folks would like to run the installation in a single step, so no async and retries if not needed (some conditional that if the `increase_bootstrap_timeout` and `increase_install_timeout` variables are 0, then run the installation in sync mode, otherwise have a block to run the async mode).
> > My 2 cents.
> 
> I disagree. Part of the benefit of running the main installer async and then the wait-for's separately is that the user gets at least a little more incremental feedback from Ansible. Currently, the installer task can take up to an hour (maybe longer, depending on release version) with zero feedback on the status. By reporting the bootstrap-complete state we get intermediate feedback, even if we don't do an re-runs of the tasks.

I see the benefit, but I still prefer to have the chance to do it in a single step as if I were a user I didn't care about the installation steps.
Either way, it is just me being picky so I'll let you decide :)

sreichar:
I have to side with @e-minguez ,   and agree that adding the option for multiple waits is goof.

If we support a single flow install, we need the ability to make sure we validate this method.

jobbler:
I like having this ability. It has helped me in troubleshooting an issue I am currently having. It was looking like my issue was later in the process than it is. Breaking into two steps allowed me to see the issue is in the bootstrap creation process.
Also, my issue is a timing issue. Even though the playbook fails, my cluster will come up after a while. The ability to add a longer wait time for failure is a great addition since.
My 2 cents.

rlopez133:
After taking the feedback from everyone this is what I gathered.
- Edu and Steve - want to see single steps be the default way of deployment
- Dustin and John - like getting additional feedback during the install process of potentially what went wrong. 

Regarding what is the default, I don't think I will please either camp with whatever decision is made. However, I think the compromise is included the async and looping for wait-for cmds as a troubleshooting option in the playbook. Essentially if certain timeout values are set in the installer, use the async method, if not use the single step method. 

@sreichar @jobbler @dustinblack @e-minguez -- with this compromise you both get the ability to do either option and we default to single step.

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -1,2 +1,4 @@
 ---
 # defaults file for installer
+increase_bootstrap_timeout: 0
+increase_install_timeout: 0
```
rlopez133: No need for default values as I think the goal will be to add new tasks that check if these timeouts are defined and greater than 0

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- set_fact:
+    bootstrap_timeout: "{{ (increase_bootstrap_timeout + 1) }}"
+
+- set_fact:
+    install_timeout: "{{ (increase_install_timeout + 1) }}"
+
```
rlopez133: Why is the +1 needed?

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -7,4 +13,23 @@
 - name: Deploy OpenShift Cluster
   shell: |
     /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug create cluster
```
rlopez133: Lets keep this exactly the same but add a when clause:
```
when: bootstrap_timeout is undefined and install_timeout is undefined
```

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -7,4 +13,23 @@
...
+  async: 3600
+  poll: 0
+  ignore_errors: yes
+  register: installer_result
```
rlopez133: Create a new Deploy OpenShift Cluster task
```
- name: Deploy OpenShift Cluster with the wait-for command
  block:
    - name: Deploy OpenShift Cluster
      shell: |
        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug create cluster
      async: 3600
      poll: 0
      ignore_errors: yes
      register: installer_result
    - name: Wait for Bootstrap Complete
      shell: |
        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug wait-for bootstrap-complete
      register: wait_for_bootstrap_result
      until: wait_for_bootstrap_result is succeeded
      retries: "{{ bootstrap_timeout|int }}"
      delay: 1

    - name: Wait for Install Complete
      shell: |
        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug wait-for install-complete
      register: wait_for_install_result
      until: wait_for_install_result is succeeded
      retries: "{{ install_timeout|int }}"
      delay: 1
  when: increase_bootstrap_timeout is defined and increase_install_timeout is defined
```

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -7,4 +13,23 @@
...
+  retries: "{{ bootstrap_timeout|int }}"
+  delay: 1
+
+- name: Wait for Install Complete
+  shell: |
+    /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug wait-for install-complete
+  register: wait_for_install_result
+  until: wait_for_install_result is succeeded
+  retries: "{{ install_timeout|int }}"
+  delay: 1
```
rlopez133: moved all this into a block that just verifies the increase_bootstrap_timeout and increase_install_timeout are defined.

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
rlopez133: Above is adhoc code so please verify.

jobbler:
I just ran with this and recieved an error that the kubeconfig file did not exist. I looked and it did. I ran the wait-for bootstrap-complete from command line and it is currently waiting as expected. This seems like a possible timing issue between the Deploy OpenShift Cluster and Wait for Bootstrap Complete plays.

TASK [installer : Deploy OpenShift Cluster] ******************************************
Tuesday 03 March 2020  14:36:29 -0600 (0:00:02.907)       0:09:02.561 ********* 
changed: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

TASK [installer : Wait for Bootstrap Complete] ***************************************
Tuesday 03 March 2020  14:36:32 -0600 (0:00:03.739)       0:09:06.300 ********* 
FAILED - RETRYING: Wait for Bootstrap Complete (1 retries left).
fatal: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]: FAILED! => {"attempts": 1, "changed": true, "cmd": "/usr/local/bin/openshift-baremetal-install --dir /home/kni/clusterconfigs --log-level debug wait-for bootstrap-complete\n", "delta": "0:00:00.088750", "end": "2020-03-03 20:36:40.367142", "msg": "non-zero return code", "rc": 1, "start": "2020-03-03 20:36:40.278392", "stderr": "level=debug msg=\"OpenShift Installer v4.3.1\"\nlevel=debug msg=\"Built from commit 2055609f95b19322ee6cfdd0bea73399297c4a3e\"\nlevel=fatal msg=\"loading kubeconfig: stat /home/kni/clusterconfigs/auth/kubeconfig: no such file or directory\"", "stderr_lines": ["level=debug msg=\"OpenShift Installer v4.3.1\"", "level=debug msg=\"Built from commit 2055609f95b19322ee6cfdd0bea73399297c4a3e\"", "level=fatal msg=\"loading kubeconfig: stat /home/kni/clusterconfigs/auth/kubeconfig: no such file or directory\""], "stdout": "", "stdout_lines": []}

ansible-ipi-install/roles/installer/defaults/main.yml
dustinblack: What is the value of a check task vs. the defaults? That just seems like overhead to me, but maybe I'm missing something.

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
dustinblack: That is vestigial code from another approach I was considering. I'll clean it up.

dustinblack: In the user context of the config file, this is an "increase" value, so it needs to be a timeout in addition to the default, thus incrementing the values by 1 before using them in the below tasks.

ansible-ipi-install/roles/installer/defaults/main.yml
dustinblack: The "+1" `set_fact`s below leverage these defaults so that we end up with a minimum of `1` for the `wait-for` loops. That's not to say there isn't a better way to do this.

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
dustinblack: So your suggestion is to deploy as normal without the async anytime the `increase` values are undefined? I would argue that the async and separate `wait-for` tasks have value on their own without the increased timeouts since they provide more incremental feedback to the user during the install process.

dustinblack:
> After taking the feedback from everyone this is what I gathered.
> 
> * Edu and Steve - want to see single steps be the default way of deployment
> * Dustin and John - like getting additional feedback during the install process of potentially what went wrong.
> 
> Regarding what is the default, I don't think I will please either camp with whatever decision is made. However, I think the compromise is including the async and looping for wait-for cmds as a troubleshooting option in the playbook. Essentially if certain timeout values are set in the installer, use the async method, if not use the single step method.
> 
> @sreichar @jobbler @dustinblack @e-minguez -- with this compromise you both get the ability to do either option and we default to single step.

Personally, I think that adding a separate if/then logic path perpetuates the problem of overloading Ansible as a scripting language, adds more lines of code to maintain, and risks idempotence. In this case, that's relatively minor, but it's a soapbox item for me.

dustinblack:
> I just ran with this and recieved an error that the kubeconfig file did not exist. I looked and it did. I ran the wait-for bootstrap-complete from command line and it is currently waiting as expected. This seems like a possible timing issue between the Deploy OpenShift Cluster and Wait for Bootstrap Complete plays.
> 
> TASK [installer : Deploy OpenShift Cluster] ******************************************
> Tuesday 03 March 2020 14:36:29 -0600 (0:00:02.907) 0:09:02.561 *********
> changed: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]
> 
> TASK [installer : Wait for Bootstrap Complete] ***************************************
> Tuesday 03 March 2020 14:36:32 -0600 (0:00:03.739) 0:09:06.300 *********
> FAILED - RETRYING: Wait for Bootstrap Complete (1 retries left).
> fatal: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]: FAILED! => {"attempts": 1, "changed": true, "cmd": "/usr/local/bin/openshift-baremetal-install --dir /home/kni/clusterconfigs --log-level debug wait-for bootstrap-complete\n", "delta": "0:00:00.088750", "end": "2020-03-03 20:36:40.367142", "msg": "non-zero return code", "rc": 1, "start": "2020-03-03 20:36:40.278392", "stderr": "level=debug msg="OpenShift Installer v4.3.1"\nlevel=debug msg="Built from commit 2055609f95b19322ee6cfdd0bea73399297c4a3e"\nlevel=fatal msg="loading kubeconfig: stat /home/kni/clusterconfigs/auth/kubeconfig: no such file or directory"", "stderr_lines": ["level=debug msg="OpenShift Installer v4.3.1"", "level=debug msg="Built from commit 2055609f95b19322ee6cfdd0bea73399297c4a3e"", "level=fatal msg="loading kubeconfig: stat /home/kni/clusterconfigs/auth/kubeconfig: no such file or directory""], "stdout": "", "stdout_lines": []}

This is a good catch. I haven't seen it, but I understand why it can happen.

I hate to solve this with a lazy `sleep` task. Let me play with adding a task to validate the kubeconfig file.

dustinblack:
> > I just ran with this and recieved an error that the kubeconfig file did not exist. I looked and it did. I ran the wait-for bootstrap-complete from command line and it is currently waiting as expected. This seems like a possible timing issue between the Deploy OpenShift Cluster and Wait for Bootstrap Complete plays.
> > TASK [installer : Deploy OpenShift Cluster] ******************************************
> > Tuesday 03 March 2020 14:36:29 -0600 (0:00:02.907) 0:09:02.561 *********
> > changed: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]
> > TASK [installer : Wait for Bootstrap Complete] ***************************************
> > Tuesday 03 March 2020 14:36:32 -0600 (0:00:03.739) 0:09:06.300 *********
> > FAILED - RETRYING: Wait for Bootstrap Complete (1 retries left).
> > fatal: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]: FAILED! => {"attempts": 1, "changed": true, "cmd": "/usr/local/bin/openshift-baremetal-install --dir /home/kni/clusterconfigs --log-level debug wait-for bootstrap-complete\n", "delta": "0:00:00.088750", "end": "2020-03-03 20:36:40.367142", "msg": "non-zero return code", "rc": 1, "start": "2020-03-03 20:36:40.278392", "stderr": "level=debug msg="OpenShift Installer v4.3.1"\nlevel=debug msg="Built from commit 2055609f95b19322ee6cfdd0bea73399297c4a3e"\nlevel=fatal msg="loading kubeconfig: stat /home/kni/clusterconfigs/auth/kubeconfig: no such file or directory"", "stderr_lines": ["level=debug msg="OpenShift Installer v4.3.1"", "level=debug msg="Built from commit 2055609f95b19322ee6cfdd0bea73399297c4a3e"", "level=fatal msg="loading kubeconfig: stat /home/kni/clusterconfigs/auth/kubeconfig: no such file or directory""], "stdout": "", "stdout_lines": []}
> 
> This is a good catch. I haven't seen it, but I understand why it can happen.
> 
> I hate to solve this with a lazy `sleep` task. Let me play with adding a task to validate the kubeconfig file.

I think the `stat` task I added should resolve this problem.

dustinblack:
Rebased on master. Maybe we could have another look at merging, @rlopez133 ?

dustinblack:
@rlopez133 @e-minguez @sreichar I've adjusted this per the recommendations in the conversation here to default to running the install task synchronously and only entering the new tasks if the extended timeout values are defined in the inventory file.

rlopez133:
> @rlopez133 @e-minguez @sreichar I've adjusted this per the recommendations in the conversation here to default to running the install task synchronously and only entering the new tasks if the extended timeout values are defined in the inventory file.

My statement still stands regarding having it optional so lgtm. I'll let the others chime in.

iranzo:
@rlopez133 speaking with Edu and the inconsistencies with the installer.

I was wondering if using jenkins uses a lower-resource environemnt that the one when triggering the script manually that could lead to those failures.

Edu suggested that getting this merged could help in creating a 'fallback' for installer failures that uses wait-for for some components to give it as a 'valid installation' even if the installer reported failure

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -9,4 +9,42 @@
...
+  block:
+    - name: Run OpenShift Cluster install as async task
+      shell: |
+        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug create cluster
+      async: 3600
+      poll: 0
+      ignore_errors: yes
+      register: installer_result
+
+    - name: Wait for kubeconfig File
```
smalleni, line 24: The right module/better fit would be wait_for https://docs.ansible.com/ansible/latest/modules/wait_for_module.html. Stat always seems to return success.

dustinblack: Indeed the stat module doesn't behave the way I expect (IMO it behaves wrongly, but that's for a different forum). Updated now to use wait_for

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -9,3 +9,38 @@
   shell: |
     /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug create cluster
   tags: install
+  when: increase_bootstrap_timeout is not defined and increase_install_timeout is not defined
```
rlopez133, line 12: @dustinblack I know this is extremely nit but would you mind making the following changes prior to merge?

replace lines 11-12 with this order
```
when:
  - increase_bootstrap_timeout is not defined
  - increase_install_timeout is not defined
tags: install
```

ansible-ipi-install/roles/installer/tasks/60_deploy_ocp.yml
```diff
@@ -9,3 +9,38 @@
...
+
+    - name: Wait for Install Complete
+      shell: |
+        /usr/local/bin/openshift-baremetal-install --dir {{ dir }} --log-level debug wait-for install-complete
+      register: wait_for_install_result
+      until: wait_for_install_result is succeeded
+      retries: "{{ increase_install_timeout|int }}"
+      delay: 1
+  tags: install
+  when: increase_bootstrap_timeout is defined or increase_install_timeout is defined
```
rlopez133, line 46: once again being super nit for lines 45-46

```
when:
  - increase_bootstrap_timeout is defined
  - increase_install_timeout is defined
tags: install
```

PR #175, opened 2020-02-25
adding the clusterosimage feature

### LINKED ISSUE #174: [RFE] Add support for `clusterOSImage` parameter in the ansible installation playbooks
It seems you can specify the RHCOS version used by the cluster with the clusterOSImage variable. See https://github.com/openshift/installer/blob/master/docs/user/metal/customization_ipi.md . This variable is only available on OCP 4.4+ so we need to set flags accordingly.

rlopez133:
# Description
adding the clusterosimage feature for 4.4+  into the install-config.yaml, also fixed grammar error in the text of boostraposimage explanation.

I don't have an environment to easily test this. I'm hoping @e-minguez and @abays can confirm

UPDATE: Confirmed clusterOSImage available in 4.3.1 GA so its not specific only to 4.4+

Fixes #174 

## Type of change

## Testing
The following tests are required:

* Confirm clusterOSImage to install-config.yaml is added when
** clusterosimage var is defined and not empty

* Confirm clusterOSImage is *not* added to install-config.yaml when
**  clusterosimage is defined but is empty string
** clusterosimage is *not* defined. 

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/templates/install-config.j2
+++ b/ansible-ipi-install/roles/node-prep/templates/install-config.j2
@@ -24,6 +24,9 @@ platform:
 {% if bootstraposimage is defined and bootstraposimage|length %}
     bootstrapOSImage: {{ bootstraposimage }}
 {% endif %}
+{% if clusterosimage is defined and clusterosimage|length) %}
+    clusterOSImage: {{ clusterosimage }}
+{% endif %}
     hosts:
 {% for host in groups['masters'] %}
       - name: {{ hostvars[host]['name'] }}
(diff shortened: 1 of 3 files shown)


ansible-ipi-install/roles/node-prep/templates/install-config.j2
```diff
@@ -23,6 +23,9 @@ platform:
 {% endif %}
 {% if bootstraposimage is defined and bootstraposimage|length %}
     bootstrapOSImage: {{ bootstraposimage }}
+{% endif %}
+{% if (release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int > 3)) and (clusterosimage is defined and clusterosimage|length) %}
```
abays: We can simplify this and leave off the version check because it turns out that 4.3 does support `clusterOSImage`.  Just flagging it here -- I know we already discussed this and I don't intend to nag!

rlopez133: will fix.

abays:
:+1: 
/lgtm

PR #184, opened 2020-03-03
Moved 40_bridge into it's own role so it can be re-used easily

p3ck:
# Description
This moves the bridge setup into it's own role so that it can be re-used easily.  In DCI we are providing an example playbook which sets up the provisionhost as a virtual machine.  In order for this to work we need to bridge the physical network on the physical host.  But we don't need to do any of the other node-prep work since that will happen inside the virtual machine.

Fixes # (issue)

## Type of change

## Testing

- [ ] I've run the new role stand alone

**Test Configuration**:

- Versions: 
- Hardware:

## Checklist


rlopez133:
Hi @p3ck , instead of going the new role route, the better way to handle this will be with tags. We should add tags. Then we can filter the playbook to run only the specified tags. Sadly when I created this playbook I didn't do this. More info here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html I do have an issue open regarding creating tags so its on my radar.

p3ck:
@rlopez133 ok, sounds like a reasonable approach. Feel free to close this then.

PR #190, opened 2020-03-04
adding proxy steps to readme

### LINKED ISSUE #189: [RFE] Using a proxy with the Ansible playbook
Currently, if you attempt to use a proxy with the Ansible playbook it will fail. Add documentation as to how to handle with using http_proxy. Investigation if we will add this inside the playbook itself underway.

rlopez133:
# Description

Currently, if you attempt to use a proxy with the Ansible playbook it will fail. Add documentation as to how to handle with using http_proxy.

Fixes #189 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -609,6 +609,28 @@ will modify the home directory to that of the root user (i.e. `/root`) instead
 of using the home directory of your privileged user, `kni` with a home dir of
 `/home/kni`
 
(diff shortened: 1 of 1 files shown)


rlopez133:
I need to re-think this one a bit. I found out today that httproxy requires a bit more info in the install-config.yaml so not sure if I'm going to include that here or what.

rlopez133:
PR #249 is going to add the README and capability.

PR #198, opened 2020-03-06
Install podman if required

### LINKED ISSUE #197: Using the playbooks with RHCOS cache enabled fails because podman not installed
```
TASK [installer : Start RHCOS image cache container] ***************************
Friday 06 March 2020  11:06:19 +0000 (0:00:00.250)       0:26:45.039 ********** 

fatal: [xxx.example.com]: FAILED! => {"changed": false, "msg": "Failed to find required executable podman in paths: /sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin"}
```

e-minguez:
# Description
Install podman if required by the RHCOS image caching mechanism.

Fixes #197 

## Type of change

## Testing

- [x] Tested in my own environment, before the change it fails because podman is not explicitely installed

**Test Configuration**:

- Versions: OCP 4.4 nightly
- Hardware: Bare metal

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/30_req_packages.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/30_req_packages.yml
@@ -1,8 +1,15 @@
 ---
-- name: Install required packages
-  package:
-    name: "{{ item }}"
-    state: present
-  with_items: "{{ package_list }}"
-  become: yes
+- name: Install packages
+  block:
+  - name: Create list of packages to be installed
+    set_fact:
+      package_list: "{{ package_list + cache_package_list }}"
+    when: cache_enabled == True
+
+  - name: Install required packages
+    package:
+      name: "{{ item }}"
+      state: present
+    with_items: "{{ package_list }}"
+    become: yes
   tags: packages
--- a/ansible-ipi-install/roles/node-prep/vars/main.yml
+++ b/ansible-ipi-install/roles/node-prep/vars/main.yml
@@ -16,3 +16,6 @@ package_list:
   - policycoreutils-python3
   - https://trunk.rdoproject.org/rhel8-master/deps/latest/Packages/python3-crypto-2.6.1-18.el8ost.x86_64.rpm
   - https://trunk.rdoproject.org/rhel8-master/deps/latest/Packages/python3-pyghmi-1.0.22-2.el8ost.noarch.rpm
+
+cache_package_list:
+  - podman


e-minguez:
> Thanks for catching this.
> 
> I suspect that the "disconnected install" work that is incoming might want to use podman to run the registry container as well. Is it possible that we'll want to move this package to the list of packages in the node_prep `package_list`, such that it's installed by `30_req_packages`? Both caching and disconnected install would of course be optional, but maybe it's worth just installing podman regardless?

I'd prefer to avoid installing things that are not required (like installing podman if no cache/disconnected), but I'll let @rlopez133 to decide :)

abays:
> > Thanks for catching this.
> > I suspect that the "disconnected install" work that is incoming might want to use podman to run the registry container as well. Is it possible that we'll want to move this package to the list of packages in the node_prep `package_list`, such that it's installed by `30_req_packages`? Both caching and disconnected install would of course be optional, but maybe it's worth just installing podman regardless?
> 
> I'd prefer to avoid installing things that are not required (like installing podman if no cache/disconnected), but I'll let @rlopez133 to decide :)

Yeah, it's definitely ideal not to install unneeded packages.  Maybe another option is just to move this task into `30_req_packages.yml` with a when conditional based on `cache_enabled`?  Just thinking of ways to keep all the installs in one place.

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -66,6 +66,13 @@
     clusterosimage: "http://{{ ansible_provisioning.ipv4.address }}:8080/{{ rhcos_uri }}?sha256={{ rhcos_sha256 }}"
   when: clusterosimage is not defined or clusterosimage|length < 1
 
+- name: Install required packages
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items: "podman"
+  become: yes
+
```
rlopez133: I want to move this into node-prep/tasks/30_req_packages to keep it easy to figure out where we would find this info.

under existing task in 30_req_packages lets add
```
- name: Install required packages
  package:
    name: "{{ item }}"
    state: present
  with_items: "podman"
  become: yes
  when: cache_enabled == True
```

e-minguez: Done

ansible-ipi-install/roles/node-prep/tasks/30_req_packages.yml
```diff
@@ -1,8 +1,18 @@
...
+    become: yes
+
+  - name: Install required packages when cache_enabled
+    package:
+      name: "{{ item }}"
+      state: present
+    with_items: "podman"
+    become: yes
+    when: cache_enabled == True
   tags: packages
```
rlopez133: nit i know --- but since you decided to put in a block, do we want to remove the become: yes from each task and just have it at the end of the block?

PR #201, opened 2020-03-07
Disconnected registry

### LINKED ISSUE #170: [RFE] for ansible-ipi-install: Disconnected Registry
Would like the ability to create and use a disconnected registry. Currently WIP will send PR shortly.

jobbler:
# Description

Fixes #170 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -12,6 +12,7 @@ Table of contents
       * [The Ansible `playbook.yml`](#the-ansible-playbookyml)
       * [Adding Extra Configurations to the OpenShift Installer](#adding-extra-configurations-to-the-openshift-installer)
       * [Pre-caching RHCOS Images](#pre-caching-rhcos-images)
+      * [Disconnected Registry](#disconnected-registry)
    * [Verifying Installation](#verifying-installation)
    * [Troubleshooting](#troubleshooting)
       * [Unreachable Host](#unreachable-host)
@@ -33,26 +34,26 @@ For the manual details, visit our [Installation Guide](https://github.com/opensh
 
 # Prerequisites
 
-* Best Practice Minimum Setup: 6 Physical servers (1 provision node, 3 master 
+* Best Practice Minimum Setup: 6 Physical servers (1 provision node, 3 master
 and 2 worker nodes)
 * Minimum Setup: 4 Physical servers (1 provision node, 3 master nodes)
-* Each server needs 2 NICs pre-configured. NIC1 for the private network and 
-NIC2 for the external network. NIC interface names must be identical across all nodes. 
+* Each server needs 2 NICs pre-configured. NIC1 for the private network and
+NIC2 for the external network. NIC interface names must be identical across all nodes.
 See [issue](https://github.com/openshift/installer/issues/2762)
 * Each server should have a RAID-1 configured and initialized
 * Each server must have IPMI configured
 * Each server must have DHCP setup for external NICs
 * Each server must have DNS setup for the API, wildcard applications
-* A DNS VIP is IP on the `baremetal` network is required for reservation. 
+* A DNS VIP is IP on the `baremetal` network is required for reservation.
 Reservation is done via our DHCP server (though not required).  
-* Optional - Include DNS entries for the external hostnames for each of the 
+* Optional - Include DNS entries for the external hostnames for each of the
 servers
 * Download a copy of your [Pull secret](https://cloud.redhat.com/openshift/install/metal/user-provisioned)
-* Append to the `pull-secret.txt` the [Pull secret](https://docs.google.com/document/d/1pWRtk7IbnfPo6cSDsopUMrxS22t3VJ2PuN39MJp9tHM/edit) 
+* Append to the `pull-secret.txt` the [Pull secret](https://docs.google.com/document/d/1pWRtk7IbnfPo6cSDsopUMrxS22t3VJ2PuN39MJp9tHM/edit)
 with access to `registry.svc.ci.openshift.org` and `registry.redhat.io`
 
 
-Due to the complexities of properly configuring an environment, it is 
+Due to the complexities of properly configuring an environment, it is
 recommended to review the following steps prior to running the Ansible playbook
 as without proper setup, the Ansible playbook won't work.
 
@@ -90,9 +91,9 @@ The `ansible-ipi` playbook consists of 2 main directories:
   * contains all the modifiable variables, their default values, and their definition. Some variables are empty ensuring user's give an explicit value.
   * the setting up of your provision node, master nodes, and worker nodes. Each section will require additional details (i.e. Management credentials).
 * `roles` - contains two roles: `node-prep` and `installer`. `node-prep` handles
-all the prerequisites that the provisioner node requires prior to running the 
+all the prerequisites that the provisioner node requires prior to running the
 installer. The `installer` role handles extracting the installer, setting up
-the manifests, and running the Red Hat OpenShift installation. 
+the manifests, and running the Red Hat OpenShift installation.
 
 The tree structure is shown below:
 
@@ -155,6 +156,7 @@ The tree structure is shown below:
         │   └── main.yml
         ├── templates
         │   ├── dir.xml.j2
+        │   └── pub_nic.j2
         ├── tests
         │   ├── inventory
         │   └── test.yml
@@ -165,15 +167,15 @@ The tree structure is shown below:
 
 # Running the Ansible Playbook
 
-The following are the steps to successfully run the Ansible playbook. 
+The following are the steps to successfully run the Ansible playbook.
 
 ## The `ansible.cfg` file
 
-While the `ansible.cfg` may vary upon your environment a sample is provided 
-in the repository. 
+While the `ansible.cfg` may vary upon your environment a sample is provided
+in the repository.
 
 ~~~sh
-$ cat ansible.cfg 
+$ cat ansible.cfg
 [defaults]
 inventory=./inventory
 remote_user=kni
@@ -185,9 +187,9 @@ become_method=sudo
 
 NOTE: Ensure to change the `remote_user` as deemed
 appropriate for your environment. The `remote_user` is the user previously
-created on the provision node. 
+created on the provision node.
 
-## Ansible version 
+## Ansible version
 
 Ensure that your environment is using Ansible 2.9 or greater. The following
 command can be used to verify.
@@ -203,38 +205,6 @@ ansible 2.9.1
 ~~~
 
 NOTE: The config file section should point to the path of your `ansible.cfg`
-
-## Python modules on the localhost
-The Ansible playbook takes advantage of a few python modules that are required
-to be installed on the host running the playbook. 
-
-The python modules required are:
-- dnspython (http://www.dnspython.org/) 
-- netaddr (https://github.com/netaddr/netaddr)
-
-Depending on the operating system you are running on your host and what 
-version of python your environment is using will dictate which package
-needs to be installed.
-
-On Fedora/RHEL8.x using python2 or python3
-~~~sh
-dnf install -y pythonX-dns
-dnf install -y pythonX-netaddr
-~~~
-
-NOTE: `X` is version `2` or version `3` of python. User can verify which version
-of python they are running by attempting
-~~~sh
-python --version
-python3 --version
-~~~
-
-On CentOS 7.x systems
-~~~sh
-yum install -y python-dns
-yum install -y python-netaddr
-~~~
-
 
 ## Copy local ssh key to provision node
 
@@ -252,13 +222,13 @@ NOTE: <user> should be the user previously created on the provision node
 
 The hosts file provides all the definable variables and provides a description
 of each variable. Some of the variables are explicitly left empty and **require**
-user input for the playbook to run. 
+user input for the playbook to run.
 
 The hosts file also ensure to set up all your nodes that will be used to deploy
-IPI on baremetal. There are 3 groups: `masters`, `workers`, and `provisioner`. 
-The `masters` and `workers` group collects information about the host such as 
-its name, role, user management (i.e. iDRAC) user, user management (i.e. iDRAC) 
-password, ipmi_address, ipmi_port to access the server and the provision mac 
+IPI on baremetal. There are 3 groups: `masters`, `workers`, and `provisioner`.
+The `masters` and `workers` group collects information about the host such as
+its name, role, user management (i.e. iDRAC) user, user management (i.e. iDRAC)
+password, ipmi_address, ipmi_port to access the server and the provision mac
 address (NIC1) that resides on the provisioning network.
 
 Below is a sample of the inventory/hosts file
@@ -300,12 +270,6 @@ prov_ip=172.22.0.3
 # cache.  Default is false.
 #cache_enabled=True
 
-# (Optional) Enable IPv6 addressing instead of IPv4 addressing
-#ipv6_enabled=True
-
-# (Optional) A list of clock servers to be used in chrony by the masters and workers
-#clock_servers=["pool.ntp.org","clock.redhat.com"]
-
 ######################################
 # Vars regarding install-config.yaml #
 ######################################
@@ -316,9 +280,9 @@ domain=""
 cluster=""
 # The public CIDR address, i.e. 10.1.1.0/21
 extcidrnet=""
-# An IP reserved on the baremetal network. 
+# An IP reserved on the baremetal network.
 dnsvip=""
-# An IP reserved on the baremetal network for the API endpoint. 
+# An IP reserved on the baremetal network for the API endpoint.
 # (Optional) If not set, a DNS lookup verifies that api.<clustername>.<domain> provides an IP
 #apivip=""
 # An IP reserved on the baremetal network for the Ingress endpoint.
@@ -357,19 +321,53 @@ worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_a
 # Provision Host
 [provisioner]
 provisioner.example.com
+
+# Registry Host
+#   Define a host here to create or use a local copy of the installation registry
+#   Used for disconnected installation
+# [registry_host]
+# disconnected.example.com
+
+# [registry_host:vars]
+# The following cert_* variables are needed to create the certificates
+#   when creating a disconnected registry. They are not needed to use
+#   an existing disconnected registry.
+# cert_country=MyCountry
+# cert_state=MyState
+# cert_locality=MyCity
+# cert_organization=MyCompany
+# cert_organizational_unit=MyDepartment
+
+# The port exposed on the disconnected registry host can be changed from
+# the default 5000 to something else by changing the following variable.
+# registry_port=5000
+
+# The directory the mirrored registry files are written to can be modified from teh default /opt/registry by changing the following variable.
+# registry_dir="/opt/registry"
+
+# The following two variables must be set to use an existing disconnected registry.
+#
+# Specify a file that contains extra auth tokens to include in the
+#   pull-secret if they are not already there.
+# disconnected_registry_auths_file=/home/kni/mirror_auth.json
+
+# Specify a file that contains the addition trust bundle and image
+#   content sources for the local registry. The contents of this file
+#   will be appended to the install-config.yml file.
+# disconnected_registry_mirrors_file=/home/kni/ic-appends.yml
 ~~~
 
-NOTE: The `ipmi_address` can take a fully qualified name assuming it is 
+NOTE: The `ipmi_address` can take a fully qualified name assuming it is
 resolvable.
 
 NOTE: The `ipmi_port` examples above show how a user can specify a different `ipmi_port` for each host within their inventory file. If the `ipmi_port` variable is omitted from the inventory file, the default of 623 will be used.
 
-NOTE: A detailed description of the vars under the section 
-`Vars regarding install-config.yaml` 
+NOTE: A detailed description of the vars under the section
+`Vars regarding install-config.yaml`
 may be reviewed within  [Configure the install-config and metal3-config](https://github.com/openshift-kni/baremetal-deploy/blob/master/install-steps.md#configure-the-install-config-and-metal3-config)
-if unsure how to populate. 
+if unsure how to populate.
 
-WARNING: If no `workers` are included, do not remove the workers group 
+WARNING: If no `workers` are included, do not remove the workers group
 (`[workers]`) as it is required to properly build the `install-config.yaml` file.
 
 ## The Ansible `playbook.yml`
@@ -377,7 +375,7 @@ WARNING: If no `workers` are included, do not remove the workers group
 The Ansible playbook connects to your provision host and runs through the
 `node-prep` role and the `installer` role. No modification is necessary. All
 modifications of variables may be done within the `inventory/hosts` file. A
-sample file is located in this repository under `inventory/hosts.sample`. 
+sample file is located in this repository under `inventory/hosts.sample`.
 
 Sample playbook.yml
 ~~~sh
@@ -391,17 +389,17 @@ Sample playbook.yml
 
 ## Customizing the Node Filesystems
 If you need to modify files on the node filesystems, you can augment the "fake"
-roots for the masters and workers under the 
-`roles/installer/files/customize_filesystem/{master,worker}` directories. 
+roots for the masters and workers under the
+`roles/installer/files/customize_filesystem/{master,worker}` directories.
 Any files added here will be included in the ignition config files for each
 of the machine types, leading to permanent changes to the node filesystem.
 
 NOTE: Do not place any files directly in the "fake" root -- only in subdirectories.
-Files in the root will cause the ignition process to fail. (There is a task in the 
+Files in the root will cause the ignition process to fail. (There is a task in the
 playbook to cleanup the `.gitkeep` file in the root, if it is left in place.)
 
-This will utilize the Ignition 
-[filetranspiler tool](https://github.com/ashcrow/filetranspiler/blob/master/filetranspile), 
+This will utilize the Ignition
+[filetranspiler tool](https://github.com/ashcrow/filetranspiler/blob/master/filetranspile),
 which you can read about for more information on how to use the "fake" root directories.
 
 An example of using this customization is to disable a network interface that
@@ -426,25 +424,25 @@ symbolic link to the `master` subdirectory so that changes are universal.
 ## Adding Extra Configurations to the OpenShift Installer
 Prior to the installation of Red Hat OpenShift, you may want to include
 additional configuration files to be included during the installation. The
-`installer` role handles this. 
+`installer` role handles this.
 
 In order to include the extraconfigs, ensure to place your `yaml` files within
 the `roles/installer/files/manifests` directory. All the files provided here will be
-included when the OpenShift manifests are created. 
+included when the OpenShift manifests are created.
 
-NOTE: By default this directory is empty. 
+NOTE: By default this directory is empty.
 
 ## Pre-caching RHCOS Images
 If you wish to set up a local cache of RHCOS images on your provisioning host,
-set the `cache_enabled` variable to `True` in your hosts file.  When requested, 
+set the `cache_enabled` variable to `True` in your hosts file.  When requested,
 the playbook will pre-download RHCOS images prior to actual cluster deployment.  
-It places these images in an Apache web server container on the provisioning host 
-and modifies `install-config.yaml` to instruct the bootstrap VM to download the 
+It places these images in an Apache web server container on the provisioning host
+and modifies `install-config.yaml` to instruct the bootstrap VM to download the
 images from that web server during deployment.  
 
 WARNING: If you set the `clusterosimage` and `bootstraposimage` variables, then
-`cache_enabled` will automatically be set to `False`.  Setting these variables 
-leaves the responsibility to the end user in ensuring the RHCOS images are readily 
+`cache_enabled` will automatically be set to `False`.  Setting these variables
+leaves the responsibility to the end user in ensuring the RHCOS images are readily
 available and accessible to the provision host.
 
 ## Running the `playbook.yml`
@@ -455,15 +453,63 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
 $ ansible-playbook -i inventory/hosts playbook.yml
 ~~~
 
+## Disconnected Registry
+A disconnected registry can be used to deploy the cluster.
+This registry can exist or can be created.
+
+To use a disconnected registry, set the registries host name in
+the `[registry_host]` group in the inventory file.
+
+### Creating a New Disconnected Registry
+To create a new disconnected registry, the `disconnected_registry_auths_file`
+and `disconnected_registry_mirrors_file` variables must not be set.
+
+The certificate information used to generate the host certificate must be defined.
+These variables must be defined as variables to the `registry_host` group in
+the inventory file.
+~~~
+[registry_host:vars]
+cert_country=MyCountry
+cert_state=MyState
+cert_locality=MyCity
+cert_organization=MyCompany
+cert_organizational_unit=MyDepartment
+~~~
+
+### Using an Existing Disconnected Registry
+The `disconnected_registry_auths_file` and the
+`disconnected_registry_mirrors_file` variables must be set.
+
+The `disconnected_registry_auths_file` variable should point to a file containing json data. This will be appended to the `auths` section of the pull secret.
+
+The `disconnected_registry_mirrors_file` variable should point to a file containing the `additionalTrustBundle` and `imageContentSources` for the disconnected registry.
+The file should be in the following format. Change the disconnected.example.com and port 5000 to reflect your environment.
+~~~
+additionalTrustBundle: |
+  -----BEGIN CERTIFICATE-----
+  The servers certificate should go here
+  -----END CERTIFICATE-----
+
+imageContentSources:
+- mirrors:
+  - disconnected.example.com:5000/ocp4/openshift4
+  source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+- mirrors:
+  - disconnected.example.com:5000/ocp4/openshift4
+  source: registry.svc.ci.openshift.org/ocp/release
+~~~
+
+
+
 # Verifying Installation
 Once the playbook has successfully completed, verify that your environment is
 up and running.
 
-1. Log into the provision node 
+1. Log into the provision node
 ~~~sh
 ssh kni@provisioner.example.com
 ~~~
-NOTE: `kni` user is my privileged user. 
+NOTE: `kni` user is my privileged user.
 2. Export the `kubeconfig` file located in the `~/clusterconfigs/auth` directory
 ~~~sh
 export KUBECONFIG=~/clusterconfigs/auth/kubeconfig
@@ -483,7 +529,7 @@ worker-1.openshift.example.com               Ready    worker          19h   v1.1
 # Troubleshooting
 
 The following section troubleshoots common errors that may arise when running
-the Ansible playbook. 
+the Ansible playbook.
 
 
 ## Unreachable Host
@@ -492,7 +538,7 @@ One of the most common errors is not being able to reach the provisioner host
 and seeing an error similar to
 
 ~~~sh
-$ ansible-playbook -i inventory/hosts playbook.yml 
+$ ansible-playbook -i inventory/hosts playbook.yml
 
 PLAY [IPI on Baremetal Installation Playbook] **********************************
 
@@ -503,7 +549,7 @@ PLAY RECAP *********************************************************************
 provisioner.example.com    : ok=0    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0   
 ~~~
 
-In order to solve this issue, ensure your provisioner hostname is pingable. 
+In order to solve this issue, ensure your provisioner hostname is pingable.
 
 1. The system you are currently on can `ping` the provisioner.example.com
 ~~~sh
@@ -528,7 +574,7 @@ provisioner.example.com | SUCCESS => {
 ~~~
 4. Re-run the Ansible playbook
 ~~~sh
-$ ansible-playbook -i inventory/hosts playbook.yml 
+$ ansible-playbook -i inventory/hosts playbook.yml
 ~~~
 
 ## Permission Denied Trying To Connect To Host
@@ -536,7 +582,7 @@ $ ansible-playbook -i inventory/hosts playbook.yml
 Another very common error is getting a permission denied error similar to:
 
 ~~~sh
-$ ansible-playbook -i inventory/hosts playbook.yml 
+$ ansible-playbook -i inventory/hosts playbook.yml
 
 PLAY [IPI on Baremetal Installation Playbook] *****************************************************************************************************
 
@@ -554,13 +600,13 @@ you notice closely, the Ansible playbook attempted to use my `rlopez` user inste
 of my `kni` user since my local `ansible.cfg` did not exist **AND** I had not yet
 set the `remote_user` parameter to `kni` (my privileged user).
 
-1. When working with the Ansible playbook ensure you have an `ansible.cfg` 
+1. When working with the Ansible playbook ensure you have an `ansible.cfg`
 located in the same directory as your `playbook.yml` file. The contents of the
 `ansible.cfg` should look similar to the below with the exception of changing
 your inventory path (location of `inventory` dir) and potentially your
-privileged user if not using `kni`. 
+privileged user if not using `kni`.
 ~~~sh
-$ cat ansible.cfg 
+$ cat ansible.cfg
 [defaults]
 inventory=/path/to/baremetal-deploy/ansible-ipi-install/inventory
 remote_user=kni
@@ -588,18 +634,18 @@ provisioner.example.com | SUCCESS => {
 ~~~
 4. Re-run the Ansible playbook
 ~~~sh
-$ ansible-playbook -i inventory/hosts playbook.yml 
+$ ansible-playbook -i inventory/hosts playbook.yml
 ~~~
 
 ## Dig lookup requires the python 'dnspython' library and it is not installed
 
 One of the tasks in the `node-prep` role captures your API VIP and the Ingress
-VIP of your environment using a `lookup` via `dig`. It does this 
-[DNS query using the `dnspython` library](https://docs.ansible.com/ansible/latest/plugins/lookup/dig.html). 
-This error is a little deceiving because the the `dnspython` package does 
-**not need to be installed on the remote server** (i.e. provisioner.example.com) 
-but the package must be **installed on your local host** that is running the 
-Ansible playbook. 
+VIP of your environment using a `lookup` via `dig`. It does this
+[DNS query using the `dnspython` library](https://docs.ansible.com/ansible/latest/plugins/lookup/dig.html).
+This error is a little deceiving because the the `dnspython` package does
+**not need to be installed on the remote server** (i.e. provisioner.example.com)
+but the package must be **installed on your local host** that is running the
+Ansible playbook.
 
 ~~~sh
 TASK [node-prep : fail] ************************************************************************************************************
@@ -609,37 +655,37 @@ TASK [node-prep : Verify DNS records for API VIP, Wildcard (Ingress) VIP] ******
 fatal: [provisioner.example.com]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'dig'. Error was a <class 'ansible.errors.AnsibleError'>, original message: The dig lookup requires the python 'dnspython' library and it is not installed"}
 
 PLAY RECAP *************************************************************************************************************************
-provisioner.example.com : ok=2    changed=0    unreachable=0    failed=1    skipped=3    rescued=0    ignored=0 
+provisioner.example.com : ok=2    changed=0    unreachable=0    failed=1    skipped=3    rescued=0    ignored=0
 ~~~
 
 The above issue can be fixed by simply installing `python3-dns` on your local
 system (assuming your using an OS such as Fedora, Red Hat)
 
-On a local host running Red Hat 7.x, run: 
+On a local host running Red Hat 7.x, run:
 ~~~sh
 # sudo yum install python2-dns
 ~~~
 
-On a local host running Red Hat 8.x, run: 
+On a local host running Red Hat 8.x, run:
 ~~~sh
 # sudo dnf install python3-dns
 ~~~
 
-On a local host running Fedora, run: 
+On a local host running Fedora, run:
 ~~~sh
 # sudo dnf install python3-dns
 ~~~
 
 Re-run the Ansible playbook
 ~~~sh
-$ ansible-playbook -i inventory/hosts playbook.yml 
+$ ansible-playbook -i inventory/hosts playbook.yml
 ~~~
 
 # Gotchas
 
-## Using become: yes within ansible.cfg or inside playbook.yml 
+## Using become: yes within ansible.cfg or inside playbook.yml
 
-This Ansible playbook takes advantage of the `ansible_user_dir` variable. As 
+This Ansible playbook takes advantage of the `ansible_user_dir` variable. As
 such, it is important to note that if within your `ansible.cfg` or within
 the `playbook.yml` file the privilege escalation of `become: yes` is used, this
 will modify the home directory to that of the root user (i.e. `/root`) instead
@@ -650,11 +696,11 @@ of using the home directory of your privileged user, `kni` with a home dir of
 
 As this playbook continues to grow, there may be times when it is useful
 to run specific portions of the playbook rather than running everything the
-Ansible playbook offers. 
+Ansible playbook offers.
 
 For example, a user may only want to run the networking piece of the playbook
-or create just the pull-secret.txt file, or just clean up the environment -- 
-just to name a few. 
+or create just the pull-secret.txt file, or just clean up the environment --
+just to name a few.
 
 As such the existing playbook has many tags that can be used for such purposes.
 By running the following command you can see what options are available.
@@ -694,7 +740,7 @@ To break this down further, the following is a description of each tag.
 
 * validation - this verifies that everything in your environment is set and ready
 for OpenShift deployment and sets some required internal variables. It is *not*
-recommended to ever not include this task. It is always required. 
+recommended to ever not include this task. It is always required.
 
 * subscription - subscribe via Red Hat subscription manager
 
@@ -743,8 +789,8 @@ networking
 
 ## How to use the Ansible tags
 
-The following is an example on how to use the `--tags` option. In this example, 
-we will just install the packages to the provision node. 
+The following is an example on how to use the `--tags` option. In this example,
+we will just install the packages to the provision node.
 
 Example 1
 ~~~sh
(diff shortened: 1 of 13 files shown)


ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+        mode: u=rwX,g=rX,o=rX
+        recurse: yes
+      loop:
+        - "{{ registry_dir_auth }}"
+        - "{{ registry_dir_cert }}"
+        - "{{ registry_dir_data }}"
+      become: true
+    
+    - name: Generate htpasswd entry
+      command: htpasswd -bBn dummy dummy
```
yprokule: would be great if user/password are set with variables.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+      command: podman start {{ pod_name_registry }}
+      become: true
+    
+    - name: Mirror remote registry to local
+      command:
+        oc adm release mirror
+          -a "{{ homedir }}/pullsecret.txt"
+          --from={{ release_image | quote }}
+          --to-release-image={{ local_registry | quote }}/{{ local_repo | quote }}:{{ release_version | quote }}
+          --to={{ local_registry | quote }}/{{ local_repo | quote }}
```
yprokule: Official docs mention:
`To create the installation program that is based on the content that you mirrored, extract it and pin it to the release:`
```bash
 oc adm -a ${LOCAL_SECRET_JSON} release extract --command=openshift-install "${LOCAL_REGISTRY}/${LOCAL_REPOSITORY}:${OCP_RELEASE}"
```

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+    - name: Write htpasswd file
+      copy:
+        content: '{{ htpass_entry.stdout }}'
+        dest: "{{ registry_dir_auth }}/htpasswd"
+        owner: "{{ ansible_user }}"
+        group: "{{ ansible_user }}"
+    
+    - name: Create registry auth for pullsecret
+      set_fact:
+        disconnected_auth: '  {"{{ ansible_fqdn }}:{{ registry_port }}": {"auth": "ZHVtbXk6ZHVtbXk=", "email": "{{ ansible_user }}@redhat.com" } }'
```
yprokule: Would be great to have ability to specify alternate name for registry, and if not set fallback to `ansible_fqdn`
```suggestion
        disconnected_auth: '  {"{{ registry_name|default(ansible_fqdn) }}:{{ registry_port }}": {"auth": "ZHVtbXk6ZHVtbXk=", "email": "{{ ansible_user }}@redhat.com" } }'
```

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+        force: yes
+        backup: yes
+      become: true
+    
+    
+    - name: Update the CA trust files
+      command: update-ca-trust extract
+      become: true
+    
+    - name: Create container to serve the registry
```
abays: We could use the `podman_container.py` module here that we've added to the installer role's library instead of relying on command/shell.

ansible-ipi-install/README.md
```diff
@@ -418,6 +419,53 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
...
+The `disconnected_registry_mirrors_file` variable should contain the `additionalTrustBundle` and `imageContentSources` for the disconnected registry.
+~~~
+additionalTrustBundle: |
+  -----BEGIN CERTIFICATE-----
+  The servers certificate should go here
+  -----END CERTIFICATE-----
+  
+imageContentSources:
+- mirrors:
+  - disconnected.example.org:5000/ocp4/openshift4
```
rlopez133: change this to disconnected.example.com

ansible-ipi-install/README.md
```diff
@@ -418,6 +419,53 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
...
+  source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+- mirrors:
+  - disconnected.example.org:5000/ocp4/openshift4
```
rlopez133: change this to disconnected.example.com

ansible-ipi-install/README.md
```diff
@@ -418,6 +419,53 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
...
+  source: registry.svc.ci.openshift.org/ocp/release
```
rlopez133: is this internal registry?

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
 # defaults file for installer
 cache_enabled: false
 provision_cache_store: "{{ ansible_env.HOME }}/rhcos_image_cache/"
+
+# Information used to create certificates when creating the registry.
+cert_country: MyCountry
+cert_state: MyState
+cert_locality: MyCity
+cert_organization: MyCompany
+cert_origanizational_unit: MyBusinessUnit
```
rlopez133: Do we want to set defaults for this? Should we not instead check if they have properly filled this out?

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# A file that contained extra auths to be included in the pull secret.
+disconnected_registry_auths_file:
```
rlopez133: I don't think we should set these. Instead check if the variable is defined, empty, or length == 0 .

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# Name of a file containing the additionalTrustBundle and mirrored
+#   repository settings. If this is set, then a local registry will not be created.
+#   Setting this causes the playbook to use an existing disconnected registry.
+disconnected_registry_mirrors_file:
```
rlopez133: I don't think we should set these. Instead check if the variable is defined, empty, or length == 0 .

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# The containers registry port
+registry_port_container: 5000
```
rlopez133: Since right now we hardcode opening this specific port (5000), I don't think we should offer this value as something that is changeable. Instead I think this variable should exist in installer/vars/main.yml

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# The port exposed on the server for the registry
+registry_port: "{{ registry_port_container }}"
```
rlopez133: Same comment as above: Since right now we hardcode opening this specific port (5000), I don't think we should offer this value as something that is changeable. Instead I think this variable should exist in installer/vars/main.yml

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# The following variables define the registry location
+registry_dir: /opt/registry
```
rlopez133: instead of /opt, can we change this to: {{ ansible_user_dir }}/registry

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+registry_dir_auth: "{{ registry_dir }}/auth"
+registry_dir_cert: "{{ registry_dir }}/certs"
+registry_dir_data: "{{ registry_dir }}/data"
```
rlopez133: these likely belong in installer/vars/main.yml

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# Name of the pod running as the registry.
+pod_name_registry: ocpdiscon-registry
```
rlopez133: belong in installer/vars/main.yml -- or you can just hardcode the name when you use the podman_container module

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# The information for the locally created registry
+local_registry: "{{ ansible_fqdn }}:{{ registry_port }}"
+local_repo: ocp4/openshift4
```
rlopez133: these likely belong in installer/vars/main.yml

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+
+# When a registry is created, then information to connect 
+#   to it is written to the following files.
```
rlopez133: remove 2 spaces

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,46 @@
...
+install_config_appends_file: install-config-appends.yml
+registry_auth_file: registry-auths.json
```
rlopez133: these likely belong in installer/vars/main.yml

ansible-ipi-install/roles/installer/tasks/25a_disconnected_registry_check.yml
```diff
@@ -0,0 +1,43 @@
...
+    - dra_set
+    - drm_set
+
+- name: Check if creating a new disconnected registry.
+  set_fact:
+    disconnected_registry: create
+  when:
+    - not dra_set
+    - not drm_set
+
```
rlopez133: As the way the playbook stands today, this likely belongs in the 10_validation.yml. However, I probably need to figure out how to better modularize the existing playbook. Reason I point it out is, if theres a problem it would be great if user failed fast and 10_validation is one of the first tasks to run.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: I add a task to open the correct port based upon the setting. Would it be better to remove the 5000 from the other tasks? Then the port will only open when disconnected is chosen

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: same response as above

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+    
+    - name: Information
+      debug:
+        msg:
+        - "To use this mirrored registry to deploy you must do the following"
+        - "Add the authentication from either "
+        - "    {{ homedir }}/{{ registry_auth_file }} on {{ inventory_hostname }}"
+        - "    or ~/{{ registry_auth_file }} on this server to your pull secret."
+        - ""
+        - "Append the contents of either of the "
```
abays: It looks like this is done automatically by the playbook in the `install-config.j2` template.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: i have included this and it will be pushed with my next commit

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: included. this will be pushed with my next commit

ansible-ipi-install/README.md
jobbler: done

ansible-ipi-install/README.md
jobbler: done

ansible-ipi-install/README.md
jobbler: no, it routes outside

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: i set these in case people forget. this is so it does not halt the playbook. We can set these to nothing and check that each has a value during execution.
What would you like to do?

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: this check is being done in the 25a tasks. do you just want to remove the variable from the defaults?

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: this check is being done in the 25a tasks. do you just want to remove the variable from the defaults?

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: these tasks allow the deployment of a registry on a different server. possible one acting as a disconnected registry for other systems in the future. not sure ansible_user_dir would be the best option in that case.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: i dont agree. placing in defaults allows the tasks to potentially create multiple registries on the same system.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: placing these in vars makes it difficult for the user to override from command line or inventory file, etc.

ansible-ipi-install/roles/installer/tasks/25a_disconnected_registry_check.yml
jobbler: placing this in the 10 validation playbook would make it run even when not deploying the registry. 
placing it here and only running it when creating a registry makes more sense to me.
These are checks that only need to be made when creating a registry.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: if you plan to reuse a registry without recreating it, you need the information in the files that this mentions.
Maybe changing the earlier line from reading: "To use this mirrored registry to deploy you must do the following" to "To reuse this mirrored registry for other deployments, you must do the following:"

abays: Ah, I see.  Thank you for the clarification!  But yeah, I was just confused at first and thinking that I needed to quickly modify the install-config.yaml that we generate before it was passed to openshift-install.

jobbler: the disconnected pieces set the pull secret to include the auth for the disconnected server before the install-config.yaml is created. I also modified the template to include the mirror information if it is set. I tried to make it just easily plug into the install-config.yaml process.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
abays: I also had to remove the `become: true` from this task to get the container to work properly on my RHEL 8.1 baremetal prov host.  When running as a root container, I could not reach the registry.  I tried all sorts of iptables and firewalld magic in an effort to solve the problem, but nothing was efficacious.  Running as non-root made the registry endpoint reachable (without any manual firewall changes).

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: The way I would handle this is leaving these values commented out and then within your check yml file verify if those variables are defined. In the end, I don't think anyone wants a certificate with fake info anyway.

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: Correct, I think this should just be removed from defaults. You have it in the inventory/hosts file and your 25a task should check if its defined or not.

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: Correct, I think this should just be removed from defaults. You have it in the inventory/hosts file and your 25a task should check if its defined or not.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: i just removed it and my test just finished. I will push the change up with my next commit in a bit.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: ok, i will make the change as such.

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: So I agree what should happen is:
* Remove opening 5000 by default in our firewall
* Check if disconnected is enabled, if it is enabled open the port specified by the variable registry_port_container 
* We should add a variable inside inventory/hosts.sample that states
```
# Specify the tcp port to open for the disconnected registry container. Default vaue set to 5000. Uncomment to set a different port.
#registry_port_container=5000
```

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: Let me know if you agree with comment above here since its the same for this one.

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: Thats a good point. However, we could also add this in the inventory/hosts.sample file with something like
```
# Set the disconnected registry path location. Default is /opt/registry. Uncomment to set a different value. 
#registry_dir="/opt/registry"
```

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: I don't follow. What I'm trying to say here with registry_dir_{auth,cert,data} and podman_name_registry is these aren't values that you necessarily want changed. Values that maybe changed should go in defaults/main.yml however for values that aren't to change we should place them in vars/main.yml of that role

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: user should never go in defaults/main.yml to make any mods. All mods should happen at the inventory/hosts file. So if these are values you wish for them to override they should be placed there.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: there are two vars used.
registry_port_container is the port the container uses.
registry_port is the port exposed on the host.
I think it would make more sense to modify the hosts.sample file for the registry_port variable since the registry_port_container setting will seldom change.

ansible-ipi-install/roles/installer/tasks/25a_disconnected_registry_check.yml
rlopez133: This is probably fine for now tbh. We could easily just add a when clause when registry host is set for all those checks in 10_validation and fail quicker but I think I'm going to change how that all works anyway so for now carry on

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+
+- name: Create a disconnected registry
+  block:
+    - name: Make sure needed packages are installed
+      package:
+        state: present
+        name: "{{ item }}"
+      loop:
+        - "{{ packages_registry | default([]) }}"
+      become: true
```
rlopez133: so this we will want to move into the 30_packages.yml and have a when clause of when: inventory_hostname in groups['registry_host'] . You can leave this as is for now but will want to make that change before we merge in. Also moving the packages_registry from installer/vars to node-prep/vars

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+    
+    - name: open firewall port {{ registry_port }}
+      firewalld:
+        port: "{{ registry_port }}/tcp"
+        permanent: yes
+        immediate: yes
+        state: enabled
+      become: true
```
rlopez133: same with this, we will want this in node_prep under the 70_enabled_fw_services.yml

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
rlopez133: also we want to confirm all those packages are actually needed.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+    
+    - name: query getent for homedir
+      getent:
+        database: passwd
+        key: "{{ ansible_ssh_user }}"
+    
+    - name: set homedir fact
+      set_fact:
+        homedir: "{{ getent_passwd | json_query('*[4]') | first }}"
```
rlopez133: Not sure you actually need to do this. You should be able to use 
```
{{ ansible_env.HOME }}
```

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+        state: directory
+        owner: "{{ ansible_user }}"
+        group: "{{ ansible_user }}"
+        mode: u=rwX,g=rX,o=rX
+        recurse: yes
+      loop:
+        - "{{ registry_dir_auth }}"
+        - "{{ registry_dir_cert }}"
+        - "{{ registry_dir_data }}"
+      become: true
```
rlopez133: if you set this to true, the ansible_user becomes root not kni user

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+        group: "{{ ansible_user }}"
+    
+    - name: Create registry auth for pullsecret
+      set_fact:
+        disconnected_auth: '  {"{{ ansible_fqdn }}:{{ registry_port }}": {"auth": "ZHVtbXk6ZHVtbXk=", "email": "{{ ansible_user }}@redhat.com" } }'
+  
+    - name: Write auth for disconnected to registry host
+      copy:
+        content: "{{ disconnected_auth }}"
+        dest: "{{ homedir }}/{{ registry_auth_file }}"
```
rlopez133: "{{ ansible_env.HOME }}/{{ registry_auth_file }}"

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+  
+    - name: Write auth for disconnected to localhost
+      copy:
+        content: "{{ disconnected_auth }}"
+        dest: "~/{{ registry_auth_file }}"
```
rlopez133: Just for clarity put the full path.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+      register: new_pullsecret
+  
+    - name: set pullsecret with new auth
+      set_fact:
+        pullsecret: "  {{ new_pullsecret.stdout }}"
+  
+    - name: Write pullsecret to registry host
+      copy:
+        content: '{{pullsecret}}'
+        dest: "{{ homedir }}/pullsecret.txt"
```
rlopez133: "{{ ansible_env.HOME }}/pullsecret.txt"

NOTE:
* This is different than what we already capture with a file labeled pull-secret.txt -- did you mean for this?
* I'd assume you'd want it to override existing or better yet make a copy of original prior to overwriting

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+    - name: generate ssl certs
+      shell: |
+        openssl req \
+          -newkey rsa:4096 \
+            -nodes \
+            -sha256 \
+            -keyout {{ registry_dir_cert }}/domain.key \
+            -x509 -days 365 \
+            -out {{ registry_dir_cert }}/domain.crt \
+            -subj "/C={{ cert_country }}/ST={{ cert_state }}/L={{ cert_locality }}/O={{ cert_organization }}/OU={{ cert_organizational_unit }}/CN={{ ansible_fqdn }}"
```
rlopez133: can we use this instead? https://docs.ansible.com/ansible/latest/modules/openssl_certificate_module.html

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
```diff
@@ -0,0 +1,186 @@
...
+    - name: Information
+      debug:
+        msg:
+        - "To use this mirrored registry to deploy you must do the following"
+        - "Add the authentication from either "
+        - "    {{ homedir }}/{{ registry_auth_file }} on {{ inventory_hostname }}"
+        - "    or ~/{{ registry_auth_file }} on this server to your pull secret."
+        - ""
+        - "Append the contents of either of the "
+        - "    {{ homedir }}/{{ install_config_appends_file }} on {{ inventory_hostname }} "
```
rlopez133: anywhere you used homedir, use the {{ ansible_env.HOME }}

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -1,34 +1,49 @@
...
+      tags: manifests
+    - include_tasks: 50_extramanifests.yml
+      tags: extramanifests
+    - include_tasks: 59_cleanup_bootstrap.yml
+      tags: cleanup
+    - include_tasks: 60_deploy_ocp.yml
+      tags: install
+    #- include_tasks: 70_cleanup_sub_man_registeration.yml
+    #  tags: unregister
+  when: inventory_hostname in groups['provisioner']
```
rlopez133: as we discussed, everything will run on provisioner host as that is what is specified in the playbook.yml. For the 25* tasks, just delegate_to: the registry host

ansible-ipi-install/roles/installer/vars/main.yml
```diff
@@ -3,3 +3,16 @@
...
+
+# packages needed for the disconnected registry tasks
+packages_registry:
+  - podman
+  - libvirt
+  - python3
+  - jq
+  - httpd
+  - httpd-tools
+  - python3-dns
```
rlopez133: need to reverify we need all these packages. I know libvirt, jq are already installed by default in our node-prep

rlopez133: we are also making the assumption that if they use an outside registry host, that the server has a repo setup FYI

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: Yeah that is what I meant. :)

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: i can move this to the vars dir if you want. making multiple registries on the same host probably wont be done often if at all.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: I dont expect the user to edit values in the defaults/main.yml file. Values placed in the defaults file can be overridden in the inventory file. These are set in the defaults so that the user can override them if they wish. Setting them in the vars makes it more difficult to set by the user.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: correct. but the directory these are being placed in may not be user writable. The default of /opt is the case. But the directories created under /opt are set to user writable.
Making this a become: ensure the directories can be created.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: i could not get that to generate a good certificate for me. maybe i missed something.

ansible-ipi-install/roles/installer/tasks/main.yml
jobbler: implemented this. will be in next commit

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: pushed

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: pushed

ansible-ipi-install/README.md
jobbler: pushed

ansible-ipi-install/README.md
jobbler: pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: added and pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: done and pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: done and pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: I will add something to the hosts.sample file showing this as a changeable option.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: moved these to vars
pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: moved this to vars
pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: moved to vars
pushed

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: i do that for readability. but to remain consistent with your comments, I removed the spaces.
pushed

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: changed message displayed to better reflect the intent of this.
pushed

ansible-ipi-install/roles/installer/tasks/main.yml
jobbler: pushed

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: this is a modified copy of the pull secret with the auth for the disconnected registry.
do you want me to overwrite (after backup) your existing pull-secret on disk?

ansible-ipi-install/roles/installer/vars/main.yml
jobbler: you install some of these packages on the provisioning host. If the host for the disconnected is not the provisioning host, then they will need to be installed on it.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: i was looking at this and wondering why would you need to specify a alternate name when creating the disconnected registry? The disconnected registry should have a FQDN here for the server the registry is being created on. ansible_fqdn provides this for the particular server.

I believe this can be done by changing a couple of variables, but was wondering if it was needed.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: added and will be pushed with next commit

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: done will be included in next commit.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: done will be included in next commit

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: changed this to be:
dest: "{{ lookup ('env', 'PWD') }}/{{ registry_auth_file }}"

will be included in next commit

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: done will include in next commit.

ansible-ipi-install/roles/node-prep/tasks/main.yml
```diff
@@ -1,6 +1,13 @@
 ---
 - include_tasks: 10_validation.yml
   tags: validation
+  
+- include_tasks: 15_validation_disconnected_registry.yml
+  delegate_to: localhost
```
rlopez133: this doesn't seem right. I don't think we want to delegate anything to the host running the playbook. This should delegate_to if no option given to the provisioner host.

ansible-ipi-install/roles/node-prep/vars/main.yml
```diff
@@ -19,3 +19,7 @@ package_list:
 
 cache_package_list:
   - podman
+
+# Temporary state variables for disconnected registry
+drm_set: false
+dra_set: false
```
rlopez133: These are values that can change right? If so they belong in defaults/main.yml . vars/main.yml are for vars that are pretty static that you don't want others modifying.

ansible-ipi-install/roles/node-prep/tasks/main.yml
```diff
@@ -1,6 +1,13 @@
...
+  when:
+    - "'registry_host' in groups"
+    - "groups['registry_host']"
```
rlopez133: you can just check the length of the group when: groups['registry_host'] | length == 0

rlopez133: we probably want to check if the group is defined like we have now

```
137 - fail: msg="The {{ item }} group is not defined. Please add [{{ item }}] to the inventory/hosts file"
138   when: "'{{ item }}' not in groups"
139   loop:
140     - masters
141     - workers
142   tags: validation
```
Need to add - registry_host . Sucky part about this is it will break everyone whom doesn't have this in their existing inventory file so not sure if thats the best way to proceed atm.

joeldavis84: Let me know if I'm misunderstanding your comment but the check currently is on its own and has conditionals on the existence of disconnected vars which I think probably is the way to go. If you don't have any disconnected vars defined then it seems like at that point you are pretty much straight up telling ansible that you actually do want the cluster to be connected (otherwise you would've mentioned disconnected in your config somewhere). 

The inability to omit it is probably due to the act of putting the disconnected checks in the same task as the general purpose checks and thus you can't specifically target them with a `when` to tell them to only run if they're trying to install disconnected.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: modified to extract this installer.
pushed in latest commit.

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: i added #registry_port=5000 to the hosts.sample to show user it can be changedpushed in matest commit

ansible-ipi-install/roles/installer/tasks/25a_disconnected_registry_check.yml
jobbler: i moved these validations to the node_prep role to be run as 15_* so any missing information causes a failure during the beginning of the playbook run

jobbler: in latest commit

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: per our conversation i removed the 5000 port from the 70_enabled_fw_services.yml and create them in the installer/15_disconnected_registry_create.yml.

ansible-ipi-install/roles/installer/tasks/25c_disconnected_registry_create.yml
jobbler: i moved the tasks around so that things flow a little better in with the current task files. I now rely on the installer/20_* file to create this.

jobbler: pushed in last commit

ansible-ipi-install/roles/node-prep/tasks/main.yml
jobbler: removed.
pushed in latest commit

ansible-ipi-install/roles/node-prep/vars/main.yml
jobbler: these are not meant for users to change. I put them in vars because of that. These change only when the tasks in the playbook determine they should be true.

ansible-ipi-install/roles/node-prep/tasks/main.yml
jobbler: so i can add the registry_host group to the 10_validation.yml file and modify the main.ylm as you mention. However, that will require the registry_host group to be defined even if not using a disconnected registry.
The way I have it in the main.yml allows the playbook to run if the registry_host group isnot defined. My thoughts were this makes the inventory file less complicated if disconnected is not used.
Your thoughts?

rlopez133: re-looking at this, the 
```
  when:
    - "'registry_host' in groups"
    - "groups['registry_host']"
```
as it stands is likely the way to go. I think when I commented it I overlooked that its checking if registry_host AND the group exist.

rlopez133: > so i can add the registry_host group to the 10_validation.yml file and modify the main.ylm as you mention. However, that will require the registry_host group to be defined even if not using a disconnected registry.
> The way I have it in the main.yml allows the playbook to run if the registry_host group isnot defined. My thoughts were this makes the inventory file less complicated if disconnected is not used.
> Your thoughts?

your way is better John :) Keep it as you have it.

ansible-ipi-install/README.md
```diff
@@ -418,6 +419,53 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
...
+  The servers certificate should go here
+  -----END CERTIFICATE-----
+  
+imageContentSources:
+- mirrors:
+  - disconnected.example.com:5000/ocp4/openshift4
+  source: quay.io/openshift-release-dev/ocp-v4.0-art-dev
+- mirrors:
+  - disconnected.example.com:5000/ocp4/openshift4
+  source: registry.svc.ci.openshift.org/ocp/release
```
rlopez133: These details should be in the inventory/hosts.sample file of the README.md and include comments above them (see existing examples). If the variables are part of the install-config.yaml pls put them in the `Vars regarding install-config.yaml` section of the inventory/hosts.sample file

ansible-ipi-install/README.md
```diff
@@ -418,6 +419,53 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
...
+The certificate information used to generate the host certificate must be defined.
+These variables must be defined as variables to the `registry_host` group in 
+the inventory file.
+~~~
+[registry_host:vars]
+cert_country=MyCountry
+cert_state=MyState
+cert_locality=MyCity
+cert_organization=MyCompany
+cert_organizational_unit=MyDepartment
```
rlopez133: see above comment

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -86,3 +86,37 @@ worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_a
...
+# The following two variables must be set to use an existing disconnected registry.
+#
+# Specify a file that contains extra auth tokens to include in the
+#   pull-secret if they are not already there.
+# disconnected_registry_auths_file=/home/kni/mirror_auth.json
+
+# Specify a file that contains the addition trust bundle and image
+#   content sources for the local registry. The contents of this file
+#   will be appended to the install-config.yml file.
+# disconnected_registry_mirrors_file=/home/kni/ic-appends.yml
```
rlopez133: These pieces should go under the cached_enabled=True variable in the inventory/hosts.sample

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -86,3 +86,37 @@ worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_a
...
+
+# [registry_host:vars]
+# The following cert_* variables are needed to create the certificates
+#   when creating a disconnected registry. They are not needed to use
+#   an existing disconnected registry.
+# cert_country=MyCountry
+# cert_state=MyState
+# cert_locality=MyCity
+# cert_organization=MyCompany
+# cert_organizational_unit=MyDepartment
```
rlopez133: I like this spot for these vars :)

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,21 @@
 # defaults file for installer
 cache_enabled: false
 provision_cache_store: "{{ ansible_env.HOME }}/rhcos_image_cache/"
+
+# The containers registry port
```
rlopez133: asking for opinion - should we include comments for these vars considering they are captured in the inventory/hosts.sample?

Side note: can we remove the empty line?

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133: can we remove empty line in defaults/main.yml? also see comment regarding comments in defaults/main.yml

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -2,3 +2,21 @@
...
+registry_dir: /opt/registry
+
+# When a registry is created, then information to connect 
+# to it is written to the following files.
+install_config_appends_file: install-config-appends.yml
+registry_auth_file: registry-auths.json
+
+# Default user and password for created registry
+disconnected_registry_user: dummy
+disconnected_registry_password: dummy
```
rlopez133: is this user related to the remote user we use for the ansible install? i.e. kni?

ansible-ipi-install/roles/installer/tasks/15_disconnected_registry_create.yml
```diff
@@ -0,0 +1,255 @@
...
+    - name: generate ssl certs
+      shell: |
+        openssl req \
+          -newkey rsa:4096 \
+            -nodes \
+            -sha256 \
+            -keyout {{ registry_dir_cert }}/domain.key \
+            -x509 -days 365 \
+            -out {{ registry_dir_cert }}/domain.crt \
+            -subj "/C={{ cert_country }}/ST={{ cert_state }}/L={{ cert_locality }}/O={{ cert_organization }}/OU={{ cert_organizational_unit }}/CN={{ ansible_fqdn }}"
```
rlopez133: have not tested but I think this will do what you want here:

```
- name: Generate an OpenSSL private key
  openssl_privatekey:
    path: {{ registry_dir_cert }}/domain.key
    size: 4096

- name: Generate an OpenSSL CSR
  openssl_csr:
    path: "{{ registry_dir_cert }}/domain.csr"
    privatekey_path: "{{ registry_dir_cert }}/domain.key"
    common_name: "{{ ansible_fqdn }}"
    country_name: "{{ cert_country }}"
    state_or_province_name: "{{ cert_state }}"
    locality_name: "{{ cert_locality }}"
    organization_name: "{{ cert_organization }}"
    organizational_unit_name: "{{ cert_organizational_unit }}"
    
    

- name: Generate a self signed OpenSSL Certificate
  openssl_certificate:
    path: {{ registry_dir_cert }}/domain.crt
    privatekey_path: {{ registry_dir_cert }}/domain.key
    csr_path: {{ registry_dir_cert }}/domain.csr
    provider: selfsigned
```

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -86,3 +86,37 @@ worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_a
 # Provision Host
 [provisioner]
 provisioner.example.com
+
+# Registry Host
+#   Define a host here to create or use a local copy of the installation registry
+#   Used for disconnected installation
+[registry_host]
```
rlopez133: should this line be commented out?

ansible-ipi-install/README.md
```diff
@@ -418,6 +419,53 @@ With the `playbook.yml` set and in-place, run the `playbook.yml`
...
 ~~~
 
+## Disconnected Registry
+A disconnected registry can be used to deploy the cluster.
+This registry can exist or can be created.
+
+To use a disconnected registry, set the registries host name in 
+the `[registry_host]` group in the inventory file.
+
+### Using an Existing Disconnected Registry
```
rlopez133: can we swap the order of existing disconnected and created? I will assume most people wont have one so it be great if we had that first over using existing.

ansible-ipi-install/inventory/hosts.sample
jobbler: i grouped these together with the registry_host section since they all have to do with the disconnected registry and not the cached_enabled section.
I think it would be better to leave them as is because the user could ignore them easier if they were not doing a disconnected registry. Placing var in multiple places makes it harder for the user I believe. Thoughts?

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: I guess it is a matter of preference. A short description in the file makes its a little easier when modifying the playbook. But it does clutter the file a little more.
We can remove the comments if you wish and the space as well

ansible-ipi-install/README.md
jobbler: yes, that makes sense.

ansible-ipi-install/README.md
jobbler: Would it be a little confusing to the reader if this is put there before its discussed?

ansible-ipi-install/README.md
jobbler: would this be confusing to put it there before discussing it?

ansible-ipi-install/roles/installer/defaults/main.yml
jobbler: changed will be in next commit

ansible-ipi-install/inventory/hosts.sample
jobbler: yes. will be in next commit

ansible-ipi-install/README.md
jobbler: done. will be in next commit

ansible-ipi-install/README.md
jobbler: done in next commit

ansible-ipi-install/README.md
jobbler: done in next commit

ansible-ipi-install/inventory/hosts.sample
jobbler: i just realized i need to change the way the registry_port and registry_dir vars are handled. they need references/set to the hosts[groups[registry_host][0]] method.

jobbler: fixed. in next commit

ansible-ipi-install/roles/installer/tasks/15_disconnected_registry_create.yml
jobbler: working. thanks. in next commit

jobbler:
I am submitting this PR for review.

rlopez133:
[test-ipi]

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -4,6 +4,17 @@
   tags:
   - cleanup
   - getoc
+
+- include_tasks: 15_disconnected_registry_existing.yml
+  when:
+    - disconnected_registry is defined
+    - disconnected_registry == "existing"
```
rlopez133: these should have tags. 
thinking:
```
tags:
- disconnected_registry
```

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -4,6 +4,17 @@
...
+
+- include_tasks: 15_disconnected_registry_create.yml
+  when:
+    - disconnected_registry is defined
+    - disconnected_registry == "create"
+
```
rlopez133: these should have tags.
thinking:

tags:
- disconnected_registry
I'm not sure if all your tasks within the file will need the same tag. One way to find out is to run the command

ansible-playbook -i inventory/hosts playbook.yml --list-tasks --list-tags

ansible-ipi-install/roles/node-prep/tasks/main.yml
```diff
@@ -1,6 +1,12 @@
 ---
 - include_tasks: 10_validation.yml
   tags: validation
+  
+- include_tasks: 15_validation_disconnected_registry.yml
+  when:
+    - "'registry_host' in groups"
+    - "groups['registry_host']"
```
rlopez133: needs tags

jobbler: done pushed in commit

ansible-ipi-install/roles/installer/tasks/main.yml
jobbler: done added in commit

ansible-ipi-install/roles/installer/tasks/main.yml
jobbler: done pushed in commit

PR #204, opened 2020-03-09
Get connection name for prov_nic

yprokule:
# Description

On RHEL8 there is no connection with name matching value of `prov_nic` variable.
To delete connection that uses `prov_nic` interface obtain its name by filtering output of `nmcli -f device,uuid con` command

## Type of change

## Testing

Run the playbook

**Test Configuration**:

- Versions: 4.3.0-0.nightly-2020-03-04-165955
- Hardware: vms

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/40_bridge.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/40_bridge.yml
@@ -2,11 +2,17 @@
 - name: Setup Bridge Creation
   block:
     - name: Delete {{ prov_nic }} due to modify nmcli bug
-      nmcli: 
+      nmcli:
         conn_name: "{{ prov_nic }}"
         type: ethernet
         state: absent
-        
+
+    - name: Delete System {{ prov_nic }} due to modify nmcli bug
+      nmcli:
+        conn_name: "System {{ prov_nic }}"
+        type: ethernet
+        state: absent
+
     - name: Create Bridge labeled provisioning
       nmcli:
         conn_name: provisioning
@@ -16,7 +22,7 @@
         stp: off
         ip4: 172.22.0.1/24
         state: present
-    
+
     - name: Create Bridge slave on {{ prov_nic }}
       nmcli:
         conn_name: "{{ prov_nic }}"
@@ -27,7 +33,7 @@
         master: provisioning
         autoconnect: yes
         state: present
-        
+
     - name: Create Bridge labeled baremetal
       nmcli:
         conn_name: baremetal
@@ -36,7 +42,7 @@
         autoconnect: yes
         stp: off
         state: present
-    
+
     - name: Create Bridge slave on {{ pub_nic }}
       nmcli:
         conn_name: "{{ pub_nic }}"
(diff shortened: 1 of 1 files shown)


rlopez133:
Hi @yprokule , prov_nic is referring to a variable that is set within the inventory/hosts file. A sample of that file can be found here: https://github.com/openshift-kni/baremetal-deploy/tree/master/ansible-ipi-install#modifying-the-inventoryhosts-file

yprokule:
> Hi @yprokule , prov_nic is referring to a variable that is set within the inventory/hosts file. A sample of that file can be found here: https://github.com/openshift-kni/baremetal-deploy/tree/master/ansible-ipi-install#modifying-the-inventoryhosts-file

@rlopez133 hey, the problem is that in my env `prov_nic` is set to `eth0` but there is no such connection, just `System eth0`, which uses interface `eth0`

ansible-ipi-install/roles/node-prep/tasks/40_bridge.yml
```diff
@@ -1,12 +1,17 @@
...
+    - name: Get connection name using {{ prov_nic }} interface
+      shell: |
+        nmcli -f device,uuid con | awk '/{{ prov_nic }}/ {print $NF}'
+      register: prov_nic_conn
+
     - name: Delete {{ prov_nic }} due to modify nmcli bug
-      nmcli: 
-        conn_name: "{{ prov_nic }}"
+      nmcli:
+        conn_name: "{{ prov_nic_conn.stdout }}"
```
rlopez133: Hey @yprokule , instead of doing the above Lets try just these changes see if you get the result your looking for

Within roles/node-prep/defaults/main.yml add the following variable:
```
prov_nic_ansible: ansible_{{ prov_nic }}
```
Within the roles/node-prep/tasks/40_bridge.yml change the delete task to the following:
```
    - name: Delete {{ prov_nic }} due to modify nmcli bug
      nmcli:
        conn_name: "{{ hostvars[ansible_fqdn][prov_nic_ansible]['device'] }}"
        type: ethernet
        state: absent
```

This should get the proper device name for the provisioning nic.

rlopez133: Hey @yprokule , I played with this some more tonight. The fix is simple, we just need to add an additional delete task to check for the "System <eth>" device. Basically what happens if it doesn't find it, it just gives the ok status but it if it does it changes it. Once it deletes it and it starts re-creating the the connections it will create them without the System. 

The code should look as follows:
```
---
- name: Setup Bridge Creation
  block:
    - name: Delete {{ prov_nic }} due to modify nmcli bug
      nmcli:
        conn_name: "{{ prov_nic }}"
        type: ethernet
        state: absent

    - name: Delete System {{ prov_nic }} due to modify nmcli bug
      nmcli:
        conn_name: "System {{ prov_nic }}"
        type: ethernet
        state: absent
```
I put the beginning of the file and then the task where it should go. Can you test this in your environment see if it works as you expect? If so, can you update the PR to reflect this change?

Thanks.

yprokule: @rlopez133 that works, thanks

PR #223, opened 2020-03-13
Add option to define cluster_pro_if

johnbieren:
Signed-off-by: Johnny Bieren <jbieren@redhat.com>

# Description

In the virt environment, we are using a different interface in the metal3-config than prov_nic. Allow this to be configured

## Type of change

## Testing

- [X] Ran locally

**Test Configuration**:

- Versions: rhel 8.1
- Hardware: virt

## Checklist

- [X] If a change is adding a feature, it should require a change to the README.md and the review should catch this.

### DIFF
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -245,6 +245,9 @@ prov_nic=eno1
 # The public NIC (NIC2) used on all baremetal nodes
 pub_nic=eno2
 
+# (Optional) Pass a NIC to use in the metal3-config, empty value uses prov_nic
+#cluster_pro_if=""
+
 # (Optional) Activation-key for proper setup of subscription-manager, empty value skips registration
 #activation_key=""
 
--- a/ansible-ipi-install/inventory/hosts.sample
+++ b/ansible-ipi-install/inventory/hosts.sample
@@ -10,6 +10,9 @@ prov_nic=eno1
 # The public NIC (NIC2) used on all baremetal nodes
 pub_nic=eno2
 
+# (Optional) Pass a NIC to use in the metal3-config, empty value uses prov_nic
+#cluster_pro_if=""
+
 # (Optional) Activation-key for proper setup of subscription-manager, empty value skips registration
 #activation_key=""
 
--- a/ansible-ipi-install/roles/installer/templates/metal3-config.j2
+++ b/ansible-ipi-install/roles/installer/templates/metal3-config.j2
@@ -11,11 +11,14 @@ data:
   http_port: "6180"
   ironic_endpoint: http://{{ prov_ip }}:6385/v1/
   ironic_inspector_endpoint: http://{{ prov_ip }}:5050/v1/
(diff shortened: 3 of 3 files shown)


rlopez133:
hey @johnbieren I'm going to let others chime in, but in my opinion, this is very specific to your environment and not sure its a use case that needs to be captured by the playbook? I'm wondering if a better workaround for your envs would be to just modify the roles/installer/templates/metal3-config.j2 with the specific nic that you need.

rlopez133:
cc @abays and @e-minguez for their 2 cents above

yprokule:
> hey @johnbieren I'm going to let others chime in, but in my opinion, this is very specific to your environment and not sure its a use case that needs to be captured by the playbook? I'm wondering if a better workaround for your envs would be to just modify the roles/installer/templates/metal3-config.j2 with the specific nic that you need.

@rlopez133 imho, this makes playbook more easily adoptable by others, despite environment being used

abays:
> > hey @johnbieren I'm going to let others chime in, but in my opinion, this is very specific to your environment and not sure its a use case that needs to be captured by the playbook? I'm wondering if a better workaround for your envs would be to just modify the roles/installer/templates/metal3-config.j2 with the specific nic that you need.
> 
> @rlopez133 imho, this makes playbook more easily adoptable by others, despite environment being used

This actually brings up a larger concern.  Namely, that there might not be parity between the provisioning host interface names (both provisioning and baremetal) and the cluster nodes' interface names.  I don't think it matters for `pub_nic`, as that doesn't seem to be used for anything other than the provisioning host, but it indeed would be a potential concern for `prov_nic`, because we use it for both configuring the provisioning host and for setting cluster `install-config.yaml` and `metal3-config.yaml`.

rlopez133:
> > > hey @johnbieren I'm going to let others chime in, but in my opinion, this is very specific to your environment and not sure its a use case that needs to be captured by the playbook? I'm wondering if a better workaround for your envs would be to just modify the roles/installer/templates/metal3-config.j2 with the specific nic that you need.
> > 
> > 
> > @rlopez133 imho, this makes playbook more easily adoptable by others, despite environment being used
> 
> This actually brings up a larger concern. Namely, that there might not be parity between the provisioning host interface names (both provisioning and baremetal) and the cluster nodes' interface names. I don't think it matters for `pub_nic`, as that doesn't seem to be used for anything other than the provisioning host, but it indeed would be a potential concern for `prov_nic`, because we use it for both configuring the provisioning host and for setting cluster `install-config.yaml` and `metal3-config.yaml`.

To add, @yprokule @abays @johnbieren the goal is to keep the playbook as close to best practice and what is reflected in the documentation. Not to mention, adding these corner cases also makes it more difficult to troubleshoot as we may not have an environment that can test it.

rlopez133:
@yprokule @abays @johnbieren - I'm going to close this PR since the goal is to stay as close as possible to best practices and the install-steps.md

PR #225, opened 2020-03-17
adding fail msgs when packages needed on control machine

### LINKED ISSUE #213: Provide a Fail message when certain packages are needed on control machine
There are certain packages needed on the control machine (system running Ansible playbook) in order to use some Ansible pieces like lookups and filters. However, instead of directly installing on the control machine the necessary packages, we will just give a fail message detailing what is required instead of manipulating someone's localhost.

rlopez133:
# Description

Certain packages are needed to use things like lookups or filters with Ansible playbook. This lets users know what packages they need to install without actually installing them since they happen on the localhost 

Fixes #213 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -1,3 +1,44 @@
+---
+- name: Gather facts for python version on localhost
+  setup:
+    gather_subset:
+      - facter
+  delegate_to: localhost
+  tags: validation
+
+- name:  Gather the package facts on localhost
+  package_facts:
+    manager: auto
+  delegate_to: localhost
+  tags: validation
+
+- debug:
+    msg: "{{ ansible_python_version }}"
+    verbosity: 2
+  tags: validation
+
+- fail: msg="The playbook requires {{ item }}, install {{ item }} on your localhost"
+  when: 
+    - ansible_python_version is version('3.0.0', '<')
+    - item not in ansible_facts.packages
+  loop:
+    - python2-dns
+    - python2-netaddr
+  tags: validation
+
+- fail: msg="The playbook requires {{ item }}, install {{ item }} on your localhost"
+  when:
+    - ansible_python_version is version('3.0.0', '>=')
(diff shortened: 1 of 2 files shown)


ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -1,3 +1,38 @@
...
+
+- debug:
+    msg: "{{ ansible_python_version }}"
+    verbosity: 2
+  tags: validation
+
+- fail: msg="The playbook requires python2-dns and python2-netaddr, install python2-dns python2-netaddr on your localhost"
+  when: 
+    - ansible_python_version is version('3.0.0', '<')
+    - ('python2-dns' not in ansible_facts.packages or 'python2-netaddr' not in ansible_facts.packages)
```
e-minguez: Can you use with_items to avoid this list? Also, if more packages are needed, you just need to add them to the list

ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -1,3 +1,38 @@
...
+  tags: validation
+
+- fail: msg="The playbook requires python3-dns and python3-netaddr, install python3-dns python3-netaddr on your localhost"
+  when:
+    - ansible_python_version is version('3.0.0', '>=')
+    - ('python3-dns' not in ansible_facts.packages or 'python3-netaddr' not in ansible_facts.packages)
```
e-minguez: Can you use with_items to avoid this list? Also, if more packages are needed, you just need to add them to the list

ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
rlopez133: @e-minguez - not exactly sure how to do this with with_items since i need to check if that package is in the ansible_facts.packages

rlopez133: ty @e-minguez -- made the appopriate changes

PR #229, opened 2020-03-18
fixing python module dependencies

### LINKED ISSUE #227: ansible playbooks doesn't work in CentOS7 due to python2-netaddr doesn't exist
The `python2-netaddr` package doesn't exist in CentOS7, it is `python-netaddr` so as #225 has been merged, the provision fails.

rlopez133:
# Description

This fix should use the new python_requirements_info module to ensure that the python modules are installed on the localhost before continuing with the playbook. 

Fixes #227 

## Type of change

* make sure user running ansible 2.9 bc we use modules only avail here
* tell user what python modules they need on their localhost but dont install as some people dont want ansible doing things to their localhost
* went ahead and added the tag always to the 10_validation as a lot of what is in there always needs to run no matter what tag you use
* added - name to the fail tasks

## Checklist

### DIFF
--- a/ansible-ipi-install/playbook.yml
+++ b/ansible-ipi-install/playbook.yml
@@ -1,7 +1,6 @@
 ---
 - name: IPI on Baremetal Installation Playbook
   hosts: provisioner
-  gather_facts: no
   roles:
   - node-prep
   - installer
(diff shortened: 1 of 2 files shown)


p3ck:
The python checks worked on centos7 where before we had to patch this playbook. +1

Issue #242, opened 2020-03-24
Allow user to provide extcidrnet


abays:
We currently gather and set the external network CIDR from the baremetal network when handling bridge logic in the node-prep role's 40_bridge.yml.  If the user has provided the `extcidrnet` in their host variables, it ends up being overwritten.  We should defer to the user-supplied value, if present, and otherwise fall back to the gathered data.

PR #244, opened 2020-03-24
Only set extcidrnet if not provided

abays:
# Description

We should only set the `extcidrnet` variable if the user hasn't provided it in the host variables already.  Otherwise we override what they have explicitly set.

Fixes https://github.com/openshift-kni/baremetal-deploy/issues/242

## Type of change

## Testing

1. Set `extcidrnet` in your host vars.  Execute the playbook and wait until it generates `install-config.yaml`.  Then, examine either `install-config.yaml` or `install-config.yaml.bkup` and make sure that `networking.machineCIDR` is set to the value provided for `extcidrnet`.

2. Do NOT set `extcidrnet` in your host vars.  Execute the playbook and wait until it generates `install-config.yaml`.  Then, examine either `install-config.yaml` or `install-config.yaml.bkup` and make sure that `networking.machineCIDR` is set to the CIDR of the public network on the provisioning host.

**Test Configuration**:

- Versions:
- Hardware:

## Checklist


rlopez133:
[test-ipi]

PR #249, opened 2020-03-26
Adding proxy capability to the playbook and adding proxy to OpenShift

rlopez133:
# Description

Adding the ability to use a proxy when running the Ansible playbook and include the proxy settings for the OpenShift cluster 

Fixes: #189 

## Type of change

## Testing

Tested the different combinations of using/not using the different variables http_proxy, https_proxy, no_proxy with and without an external proxy. 

**Test Configuration**:

- Versions: 4.3.8

## Checklist


rlopez133:
[test-ipi]

Issue #257, opened 2020-03-31
filetranspile.py invocation fails with "ModuleNotFoundError: No module named 'magic


smalleni:
Exact error: https://gist.githubusercontent.com/smalleni/f347622a1940c144cc5b92f20593bdb6/raw/588045dc8046e37c820f57a86034160a130fabb6/gistfile1.txt

PR #258, opened 2020-03-31
Use tagged version of filetranspiler

smalleni:
In latest master filetranspiler, we see the following error when running
filetranspile.py: import magic", "ModuleNotFoundError: No module named 'magic due to the new reuirement introduced on python package file-magic.

# Description

When running the task at https://github.com/openshift-kni/baremetal-deploy/blob/master/ansible-ipi-install/roles/installer/tasks/35_customize_filesystem.yml#L45, it fails as following https://gist.githubusercontent.com/smalleni/f347622a1940c144cc5b92f20593bdb6/raw/588045dc8046e37c820f57a86034160a130fabb6/gistfile1.txt

Fixes # (issue)
https://github.com/openshift-kni/baremetal-deploy/issues/257
## Type of change

## Testing

Getting the tagged version means file-magic is no longer needed.

- Versions: Trying to install OC 4.3.5
- Hardware: RHEL 8.1 admin host

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/35_customize_filesystem.yml
+++ b/ansible-ipi-install/roles/installer/tasks/35_customize_filesystem.yml
@@ -20,7 +20,7 @@
 
   - name: Get Filetranspiler
     get_url:
-      url: https://raw.githubusercontent.com/ashcrow/filetranspiler/master/filetranspile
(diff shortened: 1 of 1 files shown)


smalleni:
@dustinblack Makes sense. Thanks for pointing that out. Updated the URL to use the 1.1.1 release.

rlopez133:
fixes #257

PR #268, opened 2020-04-01
check to see if using yum package directly fixes not installing certain packages

### LINKED ISSUE #260: Group libvirt does not exist
Running the installation playbook fails with:

```
TASK [node-prep : include_tasks] ***********************************************
Tuesday 31 March 2020  16:31:39 +0000 (0:00:00.143)       0:03:23.031 ********* 
included: /home/jenkins/baremetal-deploy/ansible-ipi-install/roles/node-prep/tasks/50_modify_sudo_user.yml for kni1-bootstrap.example.com
TASK [node-prep : Add "{{ ansible_user }}" user to libvirt group and get ssh key setup] ***
Tuesday 31 March 2020  16:31:39 +0000 (0:00:00.260)       0:03:23.291 ********* 
fatal: [kni1-bootstrap.example.com]: FAILED! => {"changed": false, "msg": "Group libvirt does not exist"}
```

Retrying the execution seems to fix it so I guess it is a race condition or something like that.

rlopez133:
# Description

We are experiencing some odd behavior where some packages are giving false positive from the package module. Attempting to use the yum module instead to force update_cache to true. The update_cache: true parameter when installing or updating packages, which forces yum to check if the package cache is out of date and it update it if needed.

Fixes #260 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/30_req_packages.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/30_req_packages.yml
@@ -7,9 +7,9 @@
     when: cache_enabled == True
 
   - name: Install required packages
-    package:
-      name: "{{ item }}"
+    yum:
+      name: "{{ package_list }}"
       state: present
-    with_items: "{{ package_list }}"
+      update_cache: true
     become: yes
   tags: packages


rlopez133:
> Using yum forces to be rpm package system if I don't recall it wrong
> 
> Additionally having the full list of packages in one go, causes that if some package has broken dependencies it could fail for everything, while doing what it's there right now, is 'slower' as it calls it several times but gets the packages that 'work' installed

The problem we are seeing that when @e-minguez does an install on his provisoner host at times the system thinks libvirt is installed when its not and gives an OK status on it which later breaks when user is being added to the libvirt group. I'm hoping by switching to yum module and update_cache this goes away. Ideally we'd use the package module IMO but considering IPI on BM will be using RHEL provisioner figured it wouldn't be a bad change.

e-minguez:
It worked for me.

PR #280, opened 2020-04-13
Poweroff only required hosts

### LINKED ISSUE #279: [RFE] Power off hosts per host basis
Current playbook powers off the servers always https://github.com/openshift-kni/baremetal-deploy/blob/master/ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml but it would be nice if that can be customized in the inventory.
My current issue is that I'm trying to run virtual masters + bare metal workers and the virtual ones are created just before running the openshift-baremetal-install, so the poweroff task fails. If enabling per host poweroff, I can specify that masters doesn't need to be powered off but workers do.

e-minguez:
# Description

Allow to customize the hosts that are powered off

Fixes #279 

## Type of change

## Testing

None yet

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -298,15 +298,17 @@ pullsecret=""
 # The hardware_profile is used by the baremetal operator to match the hardware discovered on the host
 # See https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#baremetalhost-status
 # ipmi_port is optional for each host. 623 is the common default used if omitted
+# poweroff is optional. True or ommited (by default) indicates the playbook will power off the node before deploying OCP
+#  otherwise set it to false
 [masters]
-master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default
-master-1 name=master-1 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.2 ipmi_port=623 provision_mac=ec:f4:bb:da:32:88 hardware_profile=default
-master-2 name=master-2 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.3 ipmi_port=623 provision_mac=ec:f4:bb:da:0d:98 hardware_profile=default
+master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default poweroff=true
+master-1 name=master-1 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.2 ipmi_port=623 provision_mac=ec:f4:bb:da:32:88 hardware_profile=default poweroff=true
+master-2 name=master-2 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.3 ipmi_port=623 provision_mac=ec:f4:bb:da:0d:98 hardware_profile=default poweroff=true
 
 # Worker nodes
 [workers]
-worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown
-worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown
+worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown poweroff=true
+worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown poweroff=true
 
 # Provision Host
 [provisioner]
--- a/ansible-ipi-install/inventory/hosts.sample
+++ b/ansible-ipi-install/inventory/hosts.sample
@@ -93,15 +93,17 @@ pullsecret=""
 # The hardware_profile is used by the baremetal operator to match the hardware discovered on the host
 # See https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#baremetalhost-status
 # ipmi_port is optional for each host. 623 is the common default used if omitted
+# poweroff is optional. True or ommited (by default) indicates the playbook will power off the node before deploying OCP
+#  otherwise set it to false
 [masters]
-master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default
-master-1 name=master-1 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.2 ipmi_port=623 provision_mac=ec:f4:bb:da:32:88 hardware_profile=default
-master-2 name=master-2 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.3 ipmi_port=623 provision_mac=ec:f4:bb:da:0d:98 hardware_profile=default
+master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default poweroff=true
+master-1 name=master-1 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.2 ipmi_port=623 provision_mac=ec:f4:bb:da:32:88 hardware_profile=default poweroff=true
+master-2 name=master-2 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.3 ipmi_port=623 provision_mac=ec:f4:bb:da:0d:98 hardware_profile=default poweroff=true
 
 # Worker nodes
 [workers]
-worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown
-worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown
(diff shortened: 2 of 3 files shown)


ansible-ipi-install/README.md
```diff
@@ -298,15 +298,16 @@ pullsecret=""
 # The hardware_profile is used by the baremetal operator to match the hardware discovered on the host
 # See https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#baremetalhost-status
 # ipmi_port is optional for each host. 623 is the common default used if omitted
+# poweroff=true indicates the playbook should power off the nodes before deploying OCP
```
sreichar, line 301: Since I see this as a specialty case,  I think changes should only be needed for that specialty case.  In other words, let not make every one have to make chagnes, the default should power off the nodes, and in case where you dont want to power off nodes, you can optionally add the poweroff=false

e-minguez: done

ansible-ipi-install/README.md
```diff
@@ -298,6 +298,7 @@ pullsecret=""
 # The hardware_profile is used by the baremetal operator to match the hardware discovered on the host
 # See https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#baremetalhost-status
 # ipmi_port is optional for each host. 623 is the common default used if omitted
+# poweroff=false indicates the playbook shouldn't power off the node before deploying OCP. True by default
 [masters]
 master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default
```
rlopez133, line 303: I'd place the option on all the systems like so
```
master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default poweroff=true
```
This your code already assumes true its undefined, but at least they see it clearly in the README where it should be placed it they want to change it.

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -93,6 +93,7 @@ pullsecret=""
 # The hardware_profile is used by the baremetal operator to match the hardware discovered on the host
 # See https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#baremetalhost-status
 # ipmi_port is optional for each host. 623 is the common default used if omitted
+# poweroff=false indicates the playbook shouldn't power off the node before deploying OCP. True by default
 [masters]
 master-0 name=master-0 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.1 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:58 hardware_profile=default
```
rlopez133, line 98: same as above comment for all hosts.sample

ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
```diff
@@ -1,12 +1,24 @@
...
-    - "{{ groups['workers'] }}"
+- name: Power OFF nodes
+  block:
+    - name: Create list of hosts that are going to be powered off
+      add_host:
+        groups: poweroff_hosts
+        hostname: "{{ item }}"
+        inventory_dir: "{{ hostvars[item].inventory_dir }}"
+      when: ( hostvars[item]['poweroff'] is not defined )
+        or  ( hostvars[item]['poweroff'] == true )
```
rlopez133, line 10: can you confirm if this would work?
```
( hostvars[item]['poweroff']|bool )
```
Reason I mention it is is because this expression is validating a string instead of the boolean for (true or True)

rlopez133: To add, I wonder if this would take care of both expression?
```
( hostvars[item]['poweroff']|bool | default(True))
```

ansible-ipi-install/README.md
e-minguez: Well... See https://github.com/openshift-kni/baremetal-deploy/pull/280#discussion_r407506878
At first I made it mandatory but after @sreichar 's comment I changed it to optional (which I prefer).
You decide, mandatory or optional.

ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
e-minguez:
```
$ cat pb.yaml 
---
- name: IPI on Baremetal Installation Playbook
  hosts: localhost
  tasks:
    - name: Create list of hosts that are going to be powered off
      add_host:
        groups: poweroff_hosts
        hostname: "{{ item }}"
        inventory_dir: "{{ hostvars[item].inventory_dir }}"
      when: ( hostvars[item]['poweroff']|bool | default(True) )
      with_items:
        - "{{ groups.masters }}"
        - "{{ groups.workers }}"
    
    - name: Show hosts
      debug:
        msg: "{{ item }}"
      with_items: "{{ groups['poweroff_hosts'] }}"
```

```
$ cat inv 
[masters]
master-0 
master-1
master-2

# Worker nodes
[workers]
worker-0 poweroff=true
worker-1 poweroff=false
```

```
$ ansible-playbook -i inv pb.yaml 

PLAY [IPI on Baremetal Installation Playbook] ************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Create list of hosts that are going to be powered off] *********************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check '( hostvars[item]['poweroff']|bool | default(True) )' failed. The error was: error while evaluating conditional (( hostvars[item]['poweroff']|bool | default(True) )): 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'poweroff'\n\nThe error appears to be in '/home/edu/tmp/pb.yaml': line 5, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n    - name: Create list of hosts that are going to be powered off\n      ^ here\n"}

NO MORE HOSTS LEFT ***************************************************************************************************************************************************************************************************************************

PLAY RECAP ********************************************************************************************************
```

e-minguez:
```
---
- name: IPI on Baremetal Installation Playbook
  hosts: localhost
  tasks:
    - name: Create list of hosts that are going to be powered off
      add_host:
        groups: poweroff_hosts
        hostname: "{{ item }}"
        inventory_dir: "{{ hostvars[item].inventory_dir }}"
      when: ( hostvars[item]['poweroff'] is not defined )
        or  ( hostvars[item]['poweroff']|bool | default(True) )
      with_items:
        - "{{ groups.masters }}"
        - "{{ groups.workers }}"
    
    - name: Show hosts
      debug:
        msg: "{{ item }}"
      with_items: "{{ groups['poweroff_hosts'] }}"
```

This works.

ansible-ipi-install/README.md
rlopez133: I think it should be optional BUT still show how to use it in the README. For example, ipmi_port is optional but we show the example of were to set it in the sample inventory. If someone omits it, no problem uses the default port which we set. In your case, we would just add poweroff=true. Currently as it stands, I don't think it is clear where the option should reside without an example.

ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
rlopez133: I noticed in your inventory you gave poweroff a value for both options, if you leave it out entirely does it still give the expected behavior of setting poweroff to true with the new when conditional?

e-minguez: See masters, they don't have the parameter and they are added to the group.

ansible-ipi-install/README.md
```diff
@@ -298,15 +298,17 @@ pullsecret=""
...
+master-1 name=master-1 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.2 ipmi_port=623 provision_mac=ec:f4:bb:da:32:88 hardware_profile=default poweroff=true
+master-2 name=master-2 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.3 ipmi_port=623 provision_mac=ec:f4:bb:da:0d:98 hardware_profile=default poweroff=true
 
 # Worker nodes
 [workers]
-worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown
-worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown
+worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown poweroff=false
+worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown poweroff=false
```
rlopez133, line 311: Let's keep them to true as I don't want for someone to accidentally just copy paste and get incorrect behavior if they were not diligently reading the readme

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -93,15 +93,17 @@ pullsecret=""
...
+master-1 name=master-1 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.2 ipmi_port=623 provision_mac=ec:f4:bb:da:32:88 hardware_profile=default poweroff=true
+master-2 name=master-2 role=master ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.3 ipmi_port=623 provision_mac=ec:f4:bb:da:0d:98 hardware_profile=default poweroff=true
 
 # Worker nodes
 [workers]
-worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown
-worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown
+worker-0 name=worker-0 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.4 ipmi_port=623 provision_mac=ec:f4:bb:da:0c:18 hardware_profile=unknown poweroff=false
+worker-1 name=worker-1 role=worker ipmi_user=admin ipmi_password=password ipmi_address=192.168.1.5 ipmi_port=623 provision_mac=ec:f4:bb:da:32:28 hardware_profile=unknown poweroff=false
```
rlopez133, line 106: same as above set to true

Issue #282, opened 2020-04-13
Cleaning up cache_enabled == True statements


rlopez133:
Within the Ansible playbook, there are a few situations where
```
cache_enabled == True
```
is used. I want to clean up this up and use instead
```
cache_enabled|bool
```

PR #283, opened 2020-04-13
Adding ability to specify custom webserver caching port, fixing bool

### LINKED ISSUE #281: [RFE] Ability to customizing the webserver port used for caching
Currently, the webserver hosted for caching is set to port 8080. If for some reason this port is already in use by the host an error will occur creating the caching webserver. Instead, provide the ability to specify a port and if no port is specified use the default 8080.

rlopez133:
# Description

Adding ability to specify custom webserver caching port, fixing bool statements

Fixes #281 #282 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
+++ b/ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
@@ -49,13 +49,13 @@
 
 - name: Set bootstrap image URL override if not provided by the user
   set_fact:
-    bootstraposimage: "http://{{ ansible_provisioning.ipv4.address }}:8080/{{ rhcos_qemu_uri }}?sha256={{ rhcos_qemu_sha256_unzipped }}"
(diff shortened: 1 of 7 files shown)


rlopez133:
[test-ipi]

ansible-ipi-install/roles/installer/tasks/24_rhcos_image_cache.yml
```diff
@@ -65,7 +65,7 @@
     image: registry.centos.org/centos/httpd-24-centos7:latest
     state: started
     publish:
-      - "8080:8080"
+      - "{{ webserver_caching_port_container }}:{{ webserver_caching_port }}"
```
abays, line 68: I think this might be backwards.  The first port is the host port, while the second in the internal port on the container.

rlopez133: @abays @jobbler - easy to confirm? I honestly copied the order from the 15_disconnected_create which will need a patch if true.

jobbler: the first one is the exposed port on the server. thats the firewall port you will open. the second is the containers port.
so 80:8080 would expose port 80 on the server.

rlopez133: fixed.

PR #298, opened 2020-04-16
adding flexibility mix and match ipv4 and ipv6 bridges

### LINKED ISSUE #297: [RFE] Mix and match IPv4 and IPv6 bridge environment
Currently the ipv6_enabled flag sets both the provisioning and the baremetal network to be IPv6. However, there are scenarios where you need to have an IPv4 for one network and IPv6 for the other. Have the playbook create flexibility that when ipv6_enabled is set, a user can add an additional flag if they indeed want provisioning and/or baremetal to be set to IPv4.

rlopez133:
# Description

Add the ability to have IPv4 or IPv6 mix of bridges in an environment, not just all IPv4 or all IPv6

Fixes #297 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -257,6 +257,13 @@ prov_ip=172.22.0.3
 # (Optional) Enable IPv6 addressing instead of IPv4 addressing
 #ipv6_enabled=True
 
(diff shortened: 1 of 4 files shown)


rlopez133:
[test-ipi]

PR #300, opened 2020-04-16
Allow to specify different prov_nic for masters

### LINKED ISSUE #296: [RFE] Allow different interfaces for provisioner and masters
Currently, prov_nic is being used to create the bridges in the provisioner host as well as the `provisioningNetworkInterface` parameter in the install-config.yaml
If the provisioner host and master hosts have different interfaces, this won't work.
The idea is to have a `prov_nic` as currently but also an optional `provisioningNetworkInterface` that will get by default the same value as `prov_nic` or a custom one if defined.

https://github.com/openshift-kni/baremetal-deploy/blob/master/ansible-ipi-install/roles/node-prep/tasks/40_bridge.yml

https://github.com/openshift-kni/baremetal-deploy/blob/master/ansible-ipi-install/roles/installer/templates/install-config.j2

e-minguez:
# Description

The best practice is to have same nics for provisioner and masters, but this allows more flexibility in different scenarios (labs, virtual masters,...) where the nic for the provisioner can be different while it doesn't affect the defaults either.

Fixes #296 

## Type of change

## Testing
Try to run this in an heterogeneous environment where you have a modest provisioner (with embeded nics such as eno1) and beefy masters (with dedicated nics)

**Test Configuration**:

- Versions: latest 4.4
- Hardware: baremetal

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/templates/install-config.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config.j2
@@ -32,7 +32,7 @@ platform:
     ingressVIP: {{ ingressvip }}
     dnsVIP: {{ dnsvip }}
 {% if (release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int > 3)) %}
-    provisioningNetworkInterface: {{ prov_nic }}
+    provisioningNetworkInterface: {{ masters_prov_nic }}
 {% endif %}
 {% if bootstraposimage is defined and bootstraposimage|length %}
     bootstrapOSImage: {{ bootstraposimage }}
--- a/ansible-ipi-install/README.md
+++ b/ansible-ipi-install/README.md
@@ -278,6 +278,9 @@ dnsvip=""
 # An IP reserved on the baremetal network for the Ingress endpoint.
 # (Optional) If not set, a DNS lookup verifies that *.apps.<clustername>.<domain> provides an IP
 #ingressvip=""
+# The master hosts provisioning nic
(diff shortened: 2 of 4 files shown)


ansible-ipi-install/roles/installer/templates/install-config.j2
```diff
@@ -32,7 +32,7 @@ platform:
     ingressVIP: {{ ingressvip }}
     dnsVIP: {{ dnsvip }}
 {% if (release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int > 3)) %}
-    provisioningNetworkInterface: {{ prov_nic }}
+    provisioningNetworkInterface: {{ masters_prov_nic }}
```
rlopez133, line 35: Confirming - this only applies to masters?

e-minguez: According to the documentation[1]:

```
The name of the network interface on control plane nodes connected to the provisioning network.
```
AFAIK it is 'only' used for the metal3 pod to bind to the proper interface to listen to dhcp requests, serve images, etc.

[1] https://github.com/openshift/installer/blob/master/docs/user/metal/install_ipi.md#required-inputs

iranzo:
[test-ipi]

rlopez133:
> /assign @davidvossel

david won't be the one for this, anything related to our playbook just assign to me. I was hoping I could assign to you and you could merge in but I guess we gotta tweak that in the OWNERS file.

PR #304, opened 2020-04-18
Use hostvars[host]['inventory_hostname'] in install-config.j2

smalleni:
Currently the template uses `hostvars[host]['name']` where `name`
is a hostvar set in the inventory file. However, if some other playbooks
want to consume these playbooks/roles and instead add hosts to inventory
using add_host ansible module (https://docs.ansible.com/ansible/latest/modules/add_host_module.html),
there is not really way to set the `name' hostvar as `name` is a reserved module param
in the `add_host` module and is used to supply the hostname of the new host.

Instead of changing the inventory file hostvars for master and worker groups, this commit
changes the jinja2 template making use of those vars to use `hostvars[host]['inventory_hostname']`
as that will also evaluate to master-0, worker-1 for example.

Signed-off-by: Sai Sindhur Malleni <smalleni@redhat.com>

# Description

Currently the template install-config.j2 in the installer role uses `hostvars[host]['name']` where `name`
is a hostvar set in the inventory file. However, if some other playbooks
want to consume these playbooks/roles and instead add hosts to inventory
using add_host ansible module (https://docs.ansible.com/ansible/latest/modules/add_host_module.html),
there is not really way to set the `name' hostvar as `name` is a reserved module param
in the `add_host` module and is used to supply the hostname of the new host.

Instead of changing the inventory file hostvars for master and worker groups, this commit
changes the jinja2 template making use of those vars to use `hostvars[host]['inventory_hostname']`
as that will also evaluate to master-0, worker-1 for example.

Fixes # (issue)

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/templates/install-config.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config.j2
@@ -42,7 +42,7 @@ platform:
 {% endif %}
     hosts:
 {% for host in groups['masters'] %}
-      - name: {{ hostvars[host]['name'] }}
+      - name: {{ hostvars[host]['inventory_hostname'] }}
         role: {{ hostvars[host]['role'] }}
         bmc:
 {% if 'ipmi_port' in hostvars[host] %}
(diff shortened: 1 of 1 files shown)


rlopez133:
[test-ipi]

PR #336, opened 2020-04-29
isolating network facts into its own task file

### LINKED ISSUE #335: Provisioning_subnet doesn't get value if skipping network tag in playbook
I was hoping that by providing a different tag to the provisioning subnet pieces that this would of been solved. However, I need to move this part of the code into its own task yml file to ensure it always runs.

rlopez133:
# Description

If user uses --skip-tags=network this would cause the portion of the code that calculates your provision_subnet (as well as your externalcidr) code not to run effectively causing an error since it would not appropriately fill out the install-config.yaml

Fixes #335 

## Type of change

## Checklist


rlopez133:
[test-ipi]

PR #337, opened 2020-04-29
Take backup of ignition files used by installer

smalleni:
# Description

This commit allows for taking backups of the ignition config files used by the installer for debug laster on. This feature is similar to taking backup of the install-config.yaml.

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/55_customize_filesystem.yml
+++ b/ansible-ipi-install/roles/installer/tasks/55_customize_filesystem.yml
@@ -56,6 +56,18 @@
   when: (filesFound | json_query('results[*].matched') | sum) > 0
   tags: customfs
 
+- name: Create backup of ignition config files
+  copy:
+    src: "{{ dir }}/{{ item }}.ign"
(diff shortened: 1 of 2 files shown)


ansible-ipi-install/roles/installer/tasks/55_customize_filesystem.yml
```diff
@@ -56,6 +56,18 @@
...
+  copy:
+    src: "{{ dir }}/{{ item }}.ign"
+    dest: "{{ dir }}/{{ item }}.ign.bkup"
+    owner: "{{ ansible_user }}"
+    group: "{{ ansible_user }}"
+    mode: '0644'
+    remote_src: yes
+  with_items:
+    - "master"
+    - "worker"
```
rlopez133, line 69: no need for all this, 
on line 26 add
```
backup: yes
```
https://docs.ansible.com/ansible/latest/modules/copy_module.html

smalleni: @rlopez133  However, in that case it will only backup the original ignition configs and not the ones produced after the filetranspile task "Augment Ignition Config Files"

PR #352, opened 2020-05-04
Fix for masters_nic in 4.3

### LINKED ISSUE #351: 4.3 masters_nic not implemented
#300 was missing the metal3 configmap, so the masters nic is not used but the prov_nic.

e-minguez:
# Description

Leftover from #300 as this affects 4.3

Fixes #351

## Type of change

[x] Bug fix (non-breaking change which fixes an issue)

## Testing

- [x] Tested in 4.3 without the fix and it doesn't work.

**Test Configuration**:

- Versions:
- Hardware:

## Checklist


iranzo:
[test-ipi]

PR #381, opened 2020-05-11
Enable use of --tags cache

### LINKED ISSUE #210: Using just cache tags gives Error in jmespath.search in json_query filter plugin:\n'rhcos_json' is undefined
Attempted
```
ansible-playbook -i inventory/hosts playbook.yml --tags "cache" 
```
Got:

```
TASK [installer : Set facts for RHCOS_QEMU_SHA256 and RHCOS_SHA256] *************************************
Tuesday 10 March 2020  08:59:16 -0400 (0:00:00.569)       0:00:02.295 ********* 
fatal: [provisionhost-0]: FAILED! => {"msg": "Error in jmespath.search in json_query filter plugin:\n'rhcos_json' is undefined"}
```

Solution is to add tag cache to the 23_ file as well.

iranzo:
# Description

Allow to use ansible-installer --tags "cache"

Fixes # (issue)

Closes #210
## Type of change

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -127,14 +127,6 @@
   - always
   - validation
 
-- name: Confirm whether or not internet connectivity on provisioner host
-  uri:
-    url: https://www.redhat.com
-    status_code: [-1,200]
-    timeout: 1
-  register: the_url
-  tags: cache
-
 - name: Set release_url for development envs
   set_fact:
     release_url: "{{ (webserver_url|length) | ternary(webserver_url, 'https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview') }}"
@@ -159,6 +151,8 @@
   register: check_url
   tags:
   - validation
+  - cache
+  - extract
 
 - name: Set fact disconnected_registry_auths_file
   set_fact:
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -127,6 +127,14 @@
   - always
   - validation
 
+- name: Confirm whether or not internet connectivity on provisioner host
+  uri:
+    url: https://www.redhat.com
+    status_code: [-1,200]
+    timeout: 1
+  register: the_url
+  tags: cache
+
 - name: Set release_url for development envs
   set_fact:
     release_url: "{{ (webserver_url|length) | ternary(webserver_url, 'https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview') }}"
@@ -226,6 +234,7 @@
   tags:
   - always
   - validation
+  - cache
 
 - name: Set Fact for Release Image
   set_fact:
@@ -234,6 +243,7 @@
   tags:
   - always
   - validation
+  - cache
 
 - name: Get the release version
   debug:
--- a/ansible-ipi-install/roles/installer/tasks/20_extract_installer.yml
+++ b/ansible-ipi-install/roles/installer/tasks/20_extract_installer.yml
@@ -62,15 +62,6 @@
   tags:
   - extract
 
-- name: Confirm whether or not internet connectivity on provisioner host
-  uri:
-    url: https://www.redhat.com
-    status_code: [-1,200]
-    timeout: 1
-  register: the_url
-  tags:
-  - extract
-
 - name: Setting Fact of which ansible temp file to use
   set_fact:
     tempdir_loc: "{{ disconnected_installer | ternary(registry_host_tempdir, tempdir) }}"
@@ -84,7 +75,7 @@
   args:
     chdir: "{{ tempdir }}"
     executable: /bin/bash
-  when: (disconnected_installer|length or the_url.status == 200)
+  when: (disconnected_installer|length or check_url.status == 200)
   delegate_to: "{{ disconnected_installer | ternary(groups['registry_host'][0], groups['provisioner'][0]) }}"
   tags: extract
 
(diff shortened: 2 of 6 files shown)


iranzo:
[test-ipi]

iranzo:
Trying full install it gives:

~~~
TASK [installer : Get URL of host providing the webserver] **********************************************
Monday 11 May 2020  09:23:53 +0000 (0:00:00.128)       0:02:47.142 ************                          
fatal: [kni3-myhost.com]: FAILED! => {"msg": "The task includes an opti
on with an undefined variable. The error was: 'dict object' has no attribute 'registry_host'\n\nThe error
 appears to be in '/home/kni/baremetal-deploy/ansible-ipi-install/roles/installer/tasks/24_rhcos_image_ca
che.yml': line 28, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\n
The offending line appears to be:\n\n\n- name: Get URL of host providing the webserver\n  ^ here\n"}     
~~~

iranzo:
> Trying full install it gives:
> 
> ```
> TASK [installer : Get URL of host providing the webserver] **********************************************
> Monday 11 May 2020  09:23:53 +0000 (0:00:00.128)       0:02:47.142 ************                          
> fatal: [kni3-myhost.com]: FAILED! => {"msg": "The task includes an opti
> on with an undefined variable. The error was: 'dict object' has no attribute 'registry_host'\n\nThe error
>  appears to be in '/home/kni/baremetal-deploy/ansible-ipi-install/roles/installer/tasks/24_rhcos_image_ca
> che.yml': line 28, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\n
> The offending line appears to be:\n\n\n- name: Get URL of host providing the webserver\n  ^ here\n"}     
> ```

It was temporary failure in validating connectivity, it passes now

ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -129,6 +129,14 @@
   - always
   - validation
 
+- name: Confirm whether or not internet connectivity on provisioner host
+  uri:
+    url: https://www.redhat.com
+    status_code: [-1,200]
+    timeout: 1
+  register: the_url
```
rlopez133, line 137: so existing lines 148-155 already have this exact code but the variable is labeled "check_url". We could likely leverage that. The only issue I worry about is since the variable would be declared in the node-prep role, I'm unsure by the time it reaches the installer role if it would still be usable.

iranzo: When it was not here, and using the 'tags cache', it was failing as the variable was not defined, moving this from the other module to this one, allowed it to pass, should we reuse check_url and the_url everywhere we check connectivity?

johnbieren:
[test-ipi]

ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
rlopez133: If we tag that task with cache and then include the cache tag in the main.yml under 10_validation does it work? Versus re-including the same task somewhere else.

rlopez133:
@iranzo I'm starting to have my doubts at all about this tag. For example, if you look at the task 23_rhcos_image_paths.yml it requires the openshift-baremetal-install binary exist. We have so many dependencies here for doing cache that I'm thinking if we were to offer this correctly, we probably should just create an empty podman container that servers up content on a specfic directory (i.e. rhcos_image_cache).Then it would be up to the user to populate it with the data they need. However, trying to trace back everything that needs to happen for us to populate it automatically  would cost us more time than the value this feature would provide IMO. Obviously I'm open to discussion here.

iranzo:
> @iranzo I'm starting to have my doubts at all about this tag. For example, if you look at the task 23_rhcos_image_paths.yml it requires the openshift-baremetal-install binary exist. We have so many dependencies here for doing cache that I'm thinking if we were to offer this correctly, we probably should just create an empty podman container that servers up content on a specfic directory (i.e. rhcos_image_cache).Then it would be up to the user to populate it with the data they need. However, trying to trace back everything that needs to happen for us to populate it automatically would cost us more time than the value this feature would provide IMO. Obviously I'm open to discussion here.
> 
> To add to the fire, because there is logic in the playbook that determines if the container should reside on registry host or provisioner host based on who has online access, we also need to include tags within some of the variables that deal with registry so we can properly designate the delegate_to you see across the playbook.

What was the original use case for this tag? because as it seems not trivial (i.e. it's not a fix to existing missbehavior but a feature to be developed)

rlopez133:
> > @iranzo I'm starting to have my doubts at all about this tag. For example, if you look at the task 23_rhcos_image_paths.yml it requires the openshift-baremetal-install binary exist. We have so many dependencies here for doing cache that I'm thinking if we were to offer this correctly, we probably should just create an empty podman container that servers up content on a specfic directory (i.e. rhcos_image_cache).Then it would be up to the user to populate it with the data they need. However, trying to trace back everything that needs to happen for us to populate it automatically would cost us more time than the value this feature would provide IMO. Obviously I'm open to discussion here.
> > To add to the fire, because there is logic in the playbook that determines if the container should reside on registry host or provisioner host based on who has online access, we also need to include tags within some of the variables that deal with registry so we can properly designate the delegate_to you see across the playbook.
> 
> What was the original use case for this tag? because as it seems not trivial (i.e. it's not a fix to existing missbehavior but a feature to be developed)

Original use case was to create a webserver with the RHCOS images and have this piece be independent in a sense from having to run the whole playbook to achieve it. However, since that issue was opened, we now have disconnected registries, offline installs which was not part of the playbook at that time. While tagging in my view is a nice to have feature, I don't see how easily that will be to accomplish with how integrated some of the roles are.

rlopez133:
closing this PR as a WONTFIX due to complexities of different paths that can be taken for a deployment.

PR #396, opened 2020-05-22
Added checks for VIP's

### LINKED ISSUE #331: [RFE] The playbook should warn when the VIPs are up before deploying the cluster
I've had an issue where the authentication operator didn't finished deploying hence the installation failed (see https://bugzilla.redhat.com/show_bug.cgi?id=1821283 )
In my case, it was because the ingressVIP was up in some other node, so after powering it off, the installation finished.
My suggestion is that before running the installation (the final `openshift-baremetal-install` step), run some checks to verify that the VIPs are ok (meaning 'free') by pinging or curl'ing them.

DirectedSoul1:
# Description

- This change checks the Virtual IP(VIP's) and if its UP warns the user and terminates the deployment process.  

- Issue fixed (https://github.com/openshift-kni/baremetal-deploy/issues/331) 

Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #331 

## Type of change

## Testing

The added change is simple, in a sense this deployment process terminates by warning the user if the VIP's for the BM IPI Openshift cluster is used elsewhere.

Just run the playbook for deployment. 

```
ansible-playbook -i inventory/hosts playbook.yml
```

-  If the VIP's are being used elsewhere, this change warns the user

```
TASK [node-prep : If VIP's are being used fail with an error message] ****************************************************************************************
task path: /home/kni/new_install_apr_28/baremetal-deploy/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml:119
Thursday 21 May 2020  19:23:45 -0400 (0:00:02.560)       0:00:06.389 ********** 
fatal: [virthost.ostest.test.metalkube.org]: FAILED! => {"changed": false, "msg": "192.168.111.3 and 192.168.111.4 are being used in somewhere else, kindly off the nodes to re-run the deployment"}
```

## Checklist

- [X] If the change is a fix, it should have an issue. The review should make sure the comments state the issue (not just the number) and it should use the keywords that will close the issue on merge.
- [X] A change should not being merged unless it passes CI or there is a comment/update saying what testing was passed.
- [X] PRs should not be merged unless positively reviewed.

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
+++ b/ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
@@ -1,0 +1,20 @@
+---
+- name: Capture API VIP status
+  uri:
+    url: "{{ item }}" 
+    validate_certs: no
+  register: return_status
+  ignore_errors: yes
+  loop:
+    - https://{{ apivip }}:6443/readyz
+    - https://{{ ingressvip}}:433/readyz
+  when: ('Expected output is failed with a status code of -1 as VIP's should not be active' | warn_me()) 
+  tags: vip_status
+
+- name: Fail if Ingress VIP already exists
+  fail:
+    msg: |
+      The API VIP or Ingress VIP is still active in your cluster
+      please ensure to turn off all the nodes before deployment
+  when: return_status.results[0].status != -1 or return_statusi.results[1].status != -1
+  tags: vip_status
--- a/ansible-ipi-install/roles/installer/tasks/main.yml
+++ b/ansible-ipi-install/roles/installer/tasks/main.yml
@@ -43,6 +43,8 @@
   tags: extramanifests
 - include_tasks: 55_customize_filesystem.yml
   tags: customfs
+- include_tasks: 57_check_vip_status.yml
+  tags: vip_status
 - include_tasks: 59_cleanup_bootstrap.yml
   tags: cleanup
 - include_tasks: 60_deploy_ocp.yml


ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -118,6 +118,18 @@
...
+#
+- name: Test the connectivity for VIP's
+  command: ping -c3 {{ apivip }}
+  register: results
+  tags: validation
+
+- name: If VIP's are being used fail with an error message
+  fail:
+    msg: "{{ apivip }} and {{ ingressvip }} are being used in somewhere else, kindly turn off the nodes to re-run the deployment"
+  when: '"0% packet loss" in results.stdout'
```
iranzo, line 130: I would check the return code just and add a Wait to the ping like:

ping -c3 -W 2 {{ apivip }} so that it waits '2' for each one of those 3 comm, and use the rerturn code, if it's '1' it failed, so the VIP is down, if it's '0' fail because the ip is up.

Also, does this work for both IPv4 and IPv6?

rlopez133: Couple things 
* We probably want to have this check happen just before we do the 60_* script. Reasoning is if you did a deployment and where about to deploy again, this would fail 100% of the time because the power off task has not happened yet
* The when statement checks whether you have 0% packet loss, but since you are capturing ping 3 times its possible (though not likely) that you could have a 0% packet loss in 1st message and then all of a sudden connectivity in the other 2 messages. In other words, we need to find  a better when statement to check against IMO
* Unsure if this would work on IPv4 and IPv6 as mentioned by @iranzo

DirectedSoul1: Jut confirmed this works for ipv6 as well, will look for more use cases as well. 

```
fatal: [virthost.ostest.test.metalkube.org]: FAILED! => {"changed": false, "msg": "fd35:919d:4042::1121 and 192.168.111.4 are being used in somewhere else, kindly off the nodes to re-run the deployment"}
```

rlopez133: 10_validation changes should be reverted.

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,16 @@
...
+  tags: validation
+
+- name: Print the ping results
+  debug:
+    var: results.stdout
+    verbosity: 2
+
+- name: If VIP's are being used fail with an error message
+  fail:
+    msg: "{{ apivip }} and {{ ingressvip }} are being used in somewhere else, kindly off the nodes to re-run the deployment"
```
rlopez133, line 14: This message says its checking bot apivip and ingressvip but I only see a ping test for apivip above

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Test the connectivity for VIP's
+  command: ping -c3 -i 3 {{ apivip }}
+  register: results
+  tags: validation
```
rlopez133, line 5: I'd use a different tag here all together. i.e. tags: vip_status

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,16 @@
...
+
+- name: Print the ping results
+  debug:
```
rlopez133, line 8: this module needs a tag

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Test the connectivity for VIP's
+  command: ping -c3 -i 3 {{ apivip }} && ping -c3 -i 3 {{ ingressvip }}
```
rlopez133, line 3: you should make 2 different tasks for this and check for them separately. this could be a headache assuming both work or both fail

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: Print the ping results for {{ ingressvip }}
+  debug:
+    var: ingressvip_results.stdout
+    verbosity: 2
+  tags: vip_status
+
+- name: If VIP's are being used fail with an error message
+  fail:
+    msg: "{{ apivip }} and {{ ingressvip }} are being used in somewhere else, kindly turn-off the nodes to re-run the deployment"
```
rlopez133, line 26: this should be or

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Test the connectivity for {{ apivip }}
```
rlopez133, line 2: Can we change this to "Test for existing API VIP connectivity" -- Lets not use the variable {{ apivip }} because that will show an IP and the user may not be aware directly this is what we are testing.

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Test the connectivity for {{ apivip }} 
+  command: ping -c3 -i 3 {{ apivip }} 
+  register: apivip_results
+  tags: vip_status
+
+- name: Test the connectivity for {{ ingressvip }}
```
rlopez133, line 7: Can we change this to "Test for existing Ingress VIP connectivity" -- Lets not use the variable {{ ingressvip }} because that will show an IP and the user may not be aware directly this is what we are testing.

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,28 @@
...
+  command: ping -c3 -i 3 {{ ingressvip }}
+  register: ingressvip_results
+  tags: vip_status
+
+- name: Print the ping results for {{ apivip }}
```
rlopez133, line 12: Let's change name to "API VIP ping check"

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,28 @@
...
+  debug:
+    var: apivip_results.stdout
+    verbosity: 2
+  tags: vip_status
+
+- name: Print the ping results for {{ ingressvip }}
```
rlopez133, line 18: Let's change name to "Ingress VIP ping check"

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,28 @@
...
+  debug:
+    var: ingressvip_results.stdout
+    verbosity: 2
+  tags: vip_status
+
+- name: If VIP's are being used fail with an error message
```
rlopez133, line 24: Let's change name to "Fail if API VIP or Ingress VIP are active"

rlopez133: Side note, can you squash all commits down to 1 commit?

DirectedSoul1: I just added all the suggested changes and also squashed it. Thanks.

rlopez133: > I just added all the suggested changes and also squashed it. Thanks.

Thanks for the changes. However, the commits still show 6, we want to squash them down to 1.

rlopez133:
@e-minguez this is the patch for your VIP check, can you confirm this works as you expected too?

DirectedSoul1:
I just tested the functionality.

- Successful install with 3M nodes on HPE environment
- Used only 1 master (meaning VIPs are still in use) and issued a 
```
ansible-playbook -i inventory/hosts playbook.yml  --tags=vip_status -vv 
```
This results in deployment failure, which is expected since its being used by other nodes. 

```
TASK [installer : Fail if API VIP or Ingress VIP are active] *************************************************************************************************
task path: /home/kni/new_install_apr_28/baremetal-deploy/ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml:24
Tuesday 26 May 2020  15:42:00 -0400 (0:00:00.057)       0:00:16.702 *********** 
fatal: [virthost.ostest.test.metalkube.org]: FAILED! => {"changed": false, "msg": "API VIP or Ingress VIP are being used in somewhere else, kindly turn-off the nodes to re-run the deployment"}

PLAY RECAP ***************************************************************************************************************************************************
virthost.ostest.test.metalkube.org : ok=6    changed=2    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

rlopez133:
When testing on my environment I got the following error

```
TASK [installer : Test the connectivity for existing API VIP] *************************************************************
task path: /home/rlopez/git_projects/baremetal-deploy/ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml:2
Tuesday 26 May 2020  19:32:54 -0500 (0:00:00.188)       0:05:31.408 *********** 
fatal: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]: FAILED! => {"changed": true, "cmd": ["ping", "-c3", "-i", "3", "10.19.138.86"], "delta": "0:00:06.125621", "end": "2020-05-27 00:33:03.744565", "msg": "non-zero return code", "rc": 1, "start": "2020-05-27 00:32:57.618944", "stderr": "", "stderr_lines": [], "stdout": "PING 10.19.138.86 (10.19.138.86) 56(84) bytes of data.\nFrom 10.19.140.143 icmp_seq=1 Destination Host Unreachable\nFrom 10.19.140.143 icmp_seq=2 Destination Host Unreachable\nFrom 10.19.140.143 icmp_seq=3 Destination Host Unreachable\n\n--- 10.19.138.86 ping statistics ---\n3 packets transmitted, 0 received, +3 errors, 100% packet loss, time 62ms\npipe 2", "stdout_lines": ["PING 10.19.138.86 (10.19.138.86) 56(84) bytes of data.", "From 10.19.140.143 icmp_seq=1 Destination Host Unreachable", "From 10.19.140.143 icmp_seq=2 Destination Host Unreachable", "From 10.19.140.143 icmp_seq=3 Destination Host Unreachable", "", "--- 10.19.138.86 ping statistics ---", "3 packets transmitted, 0 received, +3 errors, 100% packet loss, time 62ms", "pipe 2"]}
```
Essentially what happened here is `command` attempted to `ping` it could not `ping` and because it could not `ping` it resulted in an error. Failure to ping results in an rc value of 1 which is what we actually want because if we can `ping` that means the connection is active and the rc value would be 0.

e-minguez:
Ping is not ok to me, the icmp traffic can be blocked and that doesn't mean it doesn't work.
In order to do it properly you need to find a way to see if the VIP is up or not. For instance, the API VIP shouldn't answer https://IP:6443, dns vip, same for 53 udp, ingress same for https://ip:443 and :80
For http/https see https://docs.ansible.com/ansible/latest/modules/uri_module.html and for dns I guess https://docs.ansible.com/ansible/latest/plugins/lookup/dig.html
You just need to figure out how to fail if it works (like if the api returns something different than connection refused or timeout).

iranzo:
> Ping is not ok to me, the icmp traffic can be blocked and that doesn't mean it doesn't work.
> In order to do it properly you need to find a way to see if the VIP is up or not. For instance, the API VIP shouldn't answer https://IP:6443, dns vip, same for 53 udp, ingress same for https://ip:443 and :80
> For http/https see https://docs.ansible.com/ansible/latest/modules/uri_module.html and for dns I guess https://docs.ansible.com/ansible/latest/plugins/lookup/dig.html
> You just need to figure out how to fail if it works (like if the api returns something different than connection refused or timeout).

Initial use case description explained problems when over scaling the environment and then reinstalling using an inventory that doesn't contain the extended (scaled) hosts that might contain the VIP (even initial case described this as a ping test).

If ICMP traffic is blocked, yes it might fail, and if it's up or not, we should no longer care, if it's in use, just stop the execution. Checking if it's up and replies or not and if it gets something will be like an extra set of features or conditions.

In normal conditions this should not happen except on this kind of scenarios (where feature testing is happening), so, do we have a proper way to ensure that ip is free? (as we don't care about service running there or not, if the IP is in use, we cannot use it and we should stop earlier the installation as it will fail)

Would something like:

- name: Check all port numbers are accessible from current host
  wait_for:
    host: {vip}
    port: "{{ item }}"
    state: started         # Port should be open
    delay: 0               # No wait before first check (sec)
    timeout: 3             # Stop checking after timeout (sec)
  ignore_errors: yes
  with_items:
   - 6443
   - 443

And with 443/80 for ingress be valid in your case?

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -45,6 +45,8 @@
   tags: customfs
 - include_tasks: 59_cleanup_bootstrap.yml
   tags: cleanup
+- include_tasks: 57_check_vip_status.yml
```
iranzo, line 48: Based on file name it looks like it should be executed before, should it be using 59_1_check_vip_status or something like that?

e-minguez:
> > Ping is not ok to me, the icmp traffic can be blocked and that doesn't mean it doesn't work.
> > In order to do it properly you need to find a way to see if the VIP is up or not. For instance, the API VIP shouldn't answer https://IP:6443, dns vip, same for 53 udp, ingress same for https://ip:443 and :80
> > For http/https see https://docs.ansible.com/ansible/latest/modules/uri_module.html and for dns I guess https://docs.ansible.com/ansible/latest/plugins/lookup/dig.html
> > You just need to figure out how to fail if it works (like if the api returns something different than connection refused or timeout).
> 
> Initial use case description explained problems when over scaling the environment and then reinstalling using an inventory that doesn't contain the extended (scaled) hosts that might contain the VIP (even initial case described this as a ping test).
> 
> If ICMP traffic is blocked, yes it might fail, and if it's up or not, we should no longer care, if it's in use, just stop the execution. 

But if it fails because of timeout, you cannot be sure if it is because the IP is available or it is because there is a firewall dropping your packages.

As the API/Ingress VIPs should be available from _almost_ anywhere, the idea is to check if there is something listening there (either good or bad responses, it is ok if it answers anything except timeout)

> Checking if it's up and replies or not and if it gets something will be like an extra set of features or conditions.

To me, it is the proper way to do it, again, ping is not enough to see if it is available or not.

> In normal conditions this should not happen except on this kind of scenarios (where feature testing is happening), so, do we have a proper way to ensure that ip is free? (as we don't care about service running there or not, if the IP is in use, we cannot use it and we should stop earlier the installation as it will fail)
> 
> Would something like:
> 
> ```yaml
> - name: Check all port numbers are accessible from current host
>   wait_for:
>     host: {vip}
>     port: "{{ item }}"
>     state: started         # Port should be open
>     delay: 0               # No wait before first check (sec)
>     timeout: 3             # Stop checking after timeout (sec)
>   ignore_errors: yes
>   with_items:
>    - 6443
>    - 443
> ```
> 
> And with 443/80 for ingress be valid in your case?

ansible-ipi-install/roles/installer/tasks/main.yml
rlopez133: actually we just want 57_ to be in front of 59_

rlopez133:
> > > Ping is not ok to me, the icmp traffic can be blocked and that doesn't mean it doesn't work.
> > > In order to do it properly you need to find a way to see if the VIP is up or not. For instance, the API VIP shouldn't answer https://IP:6443, dns vip, same for 53 udp, ingress same for https://ip:443 and :80
> > > For http/https see https://docs.ansible.com/ansible/latest/modules/uri_module.html and for dns I guess https://docs.ansible.com/ansible/latest/plugins/lookup/dig.html
> > > You just need to figure out how to fail if it works (like if the api returns something different than connection refused or timeout).
> > 
> > 
> > Initial use case description explained problems when over scaling the environment and then reinstalling using an inventory that doesn't contain the extended (scaled) hosts that might contain the VIP (even initial case described this as a ping test).
> > If ICMP traffic is blocked, yes it might fail, and if it's up or not, we should no longer care, if it's in use, just stop the execution.
> 
> But if it fails because of timeout, you cannot be sure if it is because the IP is available or it is because there is a firewall dropping your packages.
> 
> As the API/Ingress VIPs should be available from _almost_ anywhere, the idea is to check if there is something listening there (either good or bad responses, it is ok if it answers anything except timeout)
> 
> > Checking if it's up and replies or not and if it gets something will be like an extra set of features or conditions.
> 
> To me, it is the proper way to do it, again, ping is not enough to see if it is available or not.
> 
> > In normal conditions this should not happen except on this kind of scenarios (where feature testing is happening), so, do we have a proper way to ensure that ip is free? (as we don't care about service running there or not, if the IP is in use, we cannot use it and we should stop earlier the installation as it will fail)
> > Would something like:
> > ```yaml
> > - name: Check all port numbers are accessible from current host
> >   wait_for:
> >     host: {vip}
> >     port: "{{ item }}"
> >     state: started         # Port should be open
> >     delay: 0               # No wait before first check (sec)
> >     timeout: 3             # Stop checking after timeout (sec)
> >   ignore_errors: yes
> >   with_items:
> >    - 6443
> >    - 443
> > ```
> > 
> > 
> > And with 443/80 for ingress be valid in your case?

I agree with @e-minguez . `ping` probably is not sufficient. However, I don't think we need to check every point. 
@e-minguez what if we used the uri_module for both API VIP and Ingress VIP and just checked if API VIP responds to https://IP:6443 and Ingress VIP responds to https://ip:443 ? I would think this would be a sufficient test no?

DirectedSoul1:
yes as @e-minguez suggested `ping` module has its own disadvantages; @rlopez133 yeah I will better use `uri` module for testing VIP's. @iranzo thanks for the feedback.

DirectedSoul1:
The logic behind 

```
- name: check if {{ apivip }} is accessible
 wait_for:
    host: '{{ apivip }}'
    port: 6443
    state: present
    delay: 0
    timeout: 10
  tags: vip_status
```
seems to be more reasonable test compared to `ping` check.

rlopez133:
> The logic behind
> 
> ```
> - name: check if {{ apivip }} is accessible
>   wait_for:
>     host: '{{ apivip }}'
>     port: 6443
>     state: present
>     delay: 0
>     timeout: 10
>   tags: vip_status
> ```
> 
> seems to be more reasonable test compared to `ping` check.

@DirectedSoul1  `wait_for` would be used if you needed to wait for the API or Ingress VIP to come up but your not waiting for it to come up, what you need is to check if a particular URI is accessible or not. 

Take a look at: https://docs.ansible.com/ansible/latest/modules/uri_module.html . This is what you need.

ansible-ipi-install/roles/installer/tasks/main.yml
DirectedSoul1: > > The logic behind
> > ```
> > - name: check if {{ apivip }} is accessible
> >   wait_for:
> >     host: '{{ apivip }}'
> >     port: 6443
> >     state: present
> >     delay: 0
> >     timeout: 10
> >   tags: vip_status
> > ```
> > 
> > 
> > seems to be more reasonable test compared to `ping` check.
> 
> @DirectedSoul1 `wait_for` would be used if you needed to wait for the API or Ingress VIP to come up but your not waiting for it to come up, what you need is to check if a particular URI is accessible or not.
> 
> Take a look at: https://docs.ansible.com/ansible/latest/modules/uri_module.html . This is what you need.

Yeah I was trying the same when I try it with `curl http://192.168.111.3:6443 -v` I'm getting the return code as `400` instead of `200` (control plane is up right now) which results in the module as shown below 
```
- name: Check the return status of {{ apivip }}
   uri:
     url: "http://{{ apivip }}:6443"
     return_content: yes
     status_code: 200  #default 200 
     register: result
  tags: validation

```

results in(maybe due to my disconnected environment) It would default to check for status code `200`  

```
{"changed": false, "content": "Client sent an HTTP request to an HTTPS server.\n", "elapsed": 0, "msg": "Status code was 400 and not [200]: HTTP Error 400: Bad Request", "redirected": false, "status": 400, "url": "http://192.168.111.3:6443"
```

DirectedSoul1:
There are two possible approaches I took 
1. Using the `wait_for` approach as suggested by @iranzo  which works but as @rlopez133 mentioned we are not waiting for any `vip` to come up 
```
- name: check if {{ apivip }} is accessible
  wait_for:
    host: '{{ apivip }}'
    port: 6443
    state: present
    delay: 0
    timeout: 10
  tags: vip_status
```
Hence the next module is `uri` which kind of give me mixed results as we use cert `https://<api-vip>` seems more appropriate 
```
 - name: Check the return status of {{ apivip }}
  uri:
    url: "https://{{ apivip }}:6443"
    return_content: yes
    status_code: 200
    register: result
  tags: vip_status
```
Gives 
```
fatal: [virthost.ostest.test.metalkube.org]: FAILED! => {"changed": false, "content": "", "elapsed": 0, "msg": "Status code was -1 and not [200]: Request failed: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:897)>", "redirected": false, "status": -1, "url": "https://192.168.111.3:6443"}
```
What is really happening is 
```
 "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/\"",
  "reason": "Forbidden",
  "details": {
```
But as a, below user permissions I should have allowed to issue req's against `API` server. 
```
oc whoami 
system:admin
```
Here `ping` and `wait_for` work, but the proper way to check any `vip's` using `uri` module is giving `400`. This has been tested by me on two clusters where the `vip's` are UP. @e-minguez any other ideas here?

e-minguez:
```
- name: 'Prepare execution'
  hosts: localhost
  connection: local
  gather_facts: no
  tasks:
    - uri:
        url: "https://dkdjiduwp.com:6443"
      register: response
      failed_when: false
      changed_when: false

    - name: 'Response status'
      debug:
        msg: "{{ response.status }}"

    - uri:
        url: "https://redhat.com:6443"
      register: response
      failed_when: false
      changed_when: false

    - name: 'Response status'
      debug:
        msg: "{{ response.status }}"

    - uri:
        url: "https://redhat.com"
      register: response
      failed_when: false
      changed_when: false

    - name: 'Response status'
      debug:
        msg: "{{ response.status }}"
```

Produce:

```
PLAY [Prepare execution] *******************************************************

TASK [uri] *********************************************************************
ok: [localhost]

TASK [Response status] *********************************************************
ok: [localhost] => {
    "msg": "-1"
}

TASK [uri] *********************************************************************
ok: [localhost]

TASK [Response status] *********************************************************
ok: [localhost] => {
    "msg": "-1"
}

TASK [uri] *********************************************************************
ok: [localhost]

TASK [Response status] *********************************************************
ok: [localhost] => {
    "msg": "200"
}

PLAY RECAP *********************************************************************
localhost                  : ok=6    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

This is a quick example of using the uri module. Check an url and register the response. You just need a when: task after that to act upon it... if the status = -1, then it failed, meaning the IP is available, meaning you can continue. If the status != -1, then it worked, meaning the IP is not available, meaning you need to fail.

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: Check for API VIP usage
+  shell: |
+    curl -kL --write-out "%{http_code}\n" --silent --output /dev/null  https://{{ apivip }}:6443/readyz
```
rlopez133, line 4: dont use shell use get_uri :)

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,26 @@
...
+ 
+- name: Print the result of API VIP
+  debug:
+    msg: "{{ api_status.stdout }}"
+    verbosity: 2
+  tags: vip_status
+ 
+- name: Check for Ingress VIP usage
+  shell: |
+    curl -kL --write-out "%{http_code}\n" --silent --output /dev/null  https://{{ ingressvip }}:443/readyz
```
rlopez133, line 17: same dont use shell use get_uri

ansible-ipi-install/roles/installer/tasks/main.yml
```diff
@@ -43,6 +43,8 @@
   tags: extramanifests
 - include_tasks: 55_customize_filesystem.yml
   tags: customfs
+- include_tasks: 57_check_vip_status.yml
```
rlopez133, line 46: you need to go into main.yml and put this in the correct numerical order

DirectedSoul1: Just ordered the tasks as per the order

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
DirectedSoul1: sorry, shell is not the right way to use here, switching to `uri` instead :)

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
DirectedSoul1: added the modif's with `uri` module

ansible-ipi-install/roles/installer/tasks/main.yml
DirectedSoul1: Added the new logic, Thanks to @rlopez133  ; Can you guys please review the above logic if it works @e-minguez @iranzo

rlopez133: @DirectedSoul1 can you squash the commits into one? thanks

DirectedSoul1: sure @rlopez133 I will do it now.

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,20 @@
...
+    - https://{{ ingressvip}}:433/readyz
+  when: ('Expected output is failed with a status code of -1 as VIP's should not be active' | warn_me()) 
+  tags: vip_status
+
+- name: Fail if Ingress VIP already exists
+  fail:
+    msg: |
+      The API VIP or Ingress VIP is still active in your cluster
+      please ensure to turn off all the nodes before deployment
+  when: return_status.results[0].status != -1 or return_statusi.results[1].status != -1
```
e-minguez, line 19: typo: `return_statusi.results[1].status != -1` (statusi)

ansible-ipi-install/roles/installer/tasks/57_check_vip_status.yml
```diff
@@ -0,0 +1,20 @@
...
+  loop:
+    - https://{{ apivip }}:6443/readyz
+    - https://{{ ingressvip}}:433/readyz
+  when: ('Expected output is failed with a status code of -1 as VIP's should not be active' | warn_me()) 
+  tags: vip_status
+
+- name: Fail if Ingress VIP already exists
+  fail:
+    msg: |
+      The API VIP or Ingress VIP is still active in your cluster
```
e-minguez, line 17: I prefer to explicitly warn the user which is the problematic VIP. Otherwise, the user will need to perform the steps manually to check which one is the on that is still up.

rlopez133: @e-minguez When you run this code, Ansible gives you output in the "Verify if API or Ingress VIP is running" task of which is the bad one. For example:

Output when API or VIP are online (giving values for both online here)
```
TASK [Verify if API or Ingress VIP is running.] *********************************************************************************************************
Tuesday 02 June 2020  07:45:14 -0500 (0:00:07.941)       0:00:07.963 ********** 
[WARNING]: Expected output is failed with a status code of -1 as VIPs should not be active

ok: [worker-0.kni4.cloud.lab.eng.bos.redhat.com] => (item=https://api.kni4.cloud.lab.eng.bos.redhat.com:6443/readyz)
failed: [worker-0.kni4.cloud.lab.eng.bos.redhat.com] (item=https://foo.apps.kni4.cloud.lab.eng.bos.redhat.com:443/readyz) => {"ansible_loop_var": "item", "cache_control": "private, max-age=0, no-cache, no-store", "changed": false, "content": "<html>\n  <head>\n    <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n\n  <style type=\"text/css\">\n  /*!\n   * Bootstrap v3.3.5 (http://getbootstrap.com)\n   * Copyright 2011-2015 Twitter, Inc.\n   * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)\n   */\n  /*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */\n  html {\n    font-family: sans-serif;\n    -ms-text-size-adjust: 100%;\n    -webkit-text-size-adjust: 100%;\n  }\n  body {\n    margin: 0;\n  }\n  h1 {\n    font-size: 1.7em;\n    font-weight: 400;\n    line-height: 1.3;\n    margin: 0.68em 0;\n  }\n  * {\n    -webkit-box-sizing: border-box;\n    -moz-box-sizing: border-box;\n    box-sizing: border-box;\n  }\n  *:before,\n  *:after {\n    -webkit-box-sizing: border-box;\n    -moz-box-sizing: border-box;\n    box-sizing: border-box;\n  }\n  html {\n    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);\n  }\n  body {\n    font-family: \"Helvetica Neue\", Helvetica, Arial, sans-serif;\n    line-height: 1.66666667;\n    font-size: 13px;\n    color: #333333;\n    background-color: #ffffff;\n    margin: 2em 1em;\n  }\n  p {\n    margin: 0 0 10px;\n    font-size: 13px;\n  }\n  .alert.alert-info {\n    padding: 15px;\n    margin-bottom: 20px;\n    border: 1px solid transparent;\n    background-color: #f5f5f5;\n    border-color: #8b8d8f;\n    color: #363636;\n    margin-top: 30px;\n  }\n  .alert p {\n    padding-left: 35px;\n  }\n  a {\n    color: #0088ce;\n  }\n\n  ul {\n    position: relative;\n    padding-left: 51px;\n  }\n  p.info {\n    position: relative;\n    font-size: 15px;\n    margin-bottom: 10px;\n  }\n  p.info:before, p.info:after {\n    content: \"\";\n    position: absolute;\n    top: 9%;\n    left: 0;\n  }\n  p.info:before {\n    content: \"i\";\n    left: 3px;\n    width: 20px;\n    height: 20px;\n    font-family: serif;\n    font-size: 15px;\n    font-weight: bold;\n    line-height: 21px;\n    text-align: center;\n    color: #fff;\n    background: #4d5258;\n    border-radius: 16px;\n  }\n\n  @media (min-width: 768px) {\n    body {\n      margin: 4em 3em;\n    }\n    h1 {\n      font-size: 2.15em;}\n  }\n\n  </style>\n  </head>\n  <body>\n    <div>\n      <h1>Application is not available</h1>\n      <p>The application is currently not serving requests at this endpoint. It may not have been started or is still starting.</p>\n\n      <div class=\"alert alert-info\">\n        <p class=\"info\">\n          Possible reasons you are seeing this page:\n        </p>\n        <ul>\n          <li>\n            <strong>The host doesn't exist.</strong>\n            Make sure the hostname was typed correctly and that a route matching this hostname exists.\n          </li>\n          <li>\n            <strong>The host exists, but doesn't have a matching path.</strong>\n            Check if the URL path was typed correctly and that the route was created using the desired path.\n          </li>\n          <li>\n            <strong>Route and path matches, but all pods are down.</strong>\n            Make sure that the resources exposed by this route (pods, services, deployment configs, etc) have at least one pod running.\n          </li>\n        </ul>\n      </div>\n    </div>\n  </body>\n</html>\n", "content_type": "text/html", "elapsed": 0, "item": "https://foo.apps.kni4.cloud.lab.eng.bos.redhat.com:443/readyz", "msg": "Status code was 503 and not [200]: HTTP Error 503: Service Unavailable", "pragma": "no-cache", "redirected": false, "status": 503, "url": "https://foo.apps.kni4.cloud.lab.eng.bos.redhat.com:443/readyz"}
...ignoring

TASK [Fail if API VIP or Ingress VIP already exists] ****************************************************************************************************
Tuesday 02 June 2020  07:45:20 -0500 (0:00:06.287)       0:00:14.251 ********** 
fatal: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]: FAILED! => {"changed": false, "msg": "The API VIP or Ingress VIP is already running, please ensure you don't have\nany rogue servers up and running.\n"}

```

Output when API or VIP are not online

```
TASK [Verify if API or Ingress VIP is running.] *********************************************************************************************************
Tuesday 02 June 2020  07:43:54 -0500 (0:00:05.825)       0:00:05.849 ********** 
[WARNING]: Expected output is failed with a status code of -1 as VIPs should not be active

failed: [worker-0.kni4.cloud.lab.eng.bos.redhat.com] (item=https://api.kni44.cloud.lab.eng.bos.redhat.com:6443/readyz) => {"ansible_loop_var": "item", "changed": false, "content": "", "elapsed": 0, "item": "https://api.kni44.cloud.lab.eng.bos.redhat.com:6443/readyz", "msg": "Status code was -1 and not [200]: Request failed: <urlopen error [Errno -2] Name or service not known>", "redirected": false, "status": -1, "url": "https://api.kni44.cloud.lab.eng.bos.redhat.com:6443/readyz"}
failed: [worker-0.kni4.cloud.lab.eng.bos.redhat.com] (item=https://foo.apps.kni44.cloud.lab.eng.bos.redhat.com:443/readyz) => {"ansible_loop_var": "item", "changed": false, "content": "", "elapsed": 0, "item": "https://foo.apps.kni44.cloud.lab.eng.bos.redhat.com:443/readyz", "msg": "Status code was -1 and not [200]: Request failed: <urlopen error [Errno -2] Name or service not known>", "redirected": false, "status": -1, "url": "https://foo.apps.kni44.cloud.lab.eng.bos.redhat.com:443/readyz"}
...ignoring

TASK [Fail if API VIP or Ingress VIP already exists] ****************************************************************************************************
Tuesday 02 June 2020  07:44:00 -0500 (0:00:06.718)       0:00:12.567 ********** 
skipping: [worker-0.kni4.cloud.lab.eng.bos.redhat.com]

```

e-minguez: It is still confusing to me because I don't know exactly which VIP is failing on the first example.

rlopez133: @e-minguez This is why I added that WARNING because it is confusing. In fact the uri module expects error when it sees a value of `-1` and we see success when we see `-1`. This is the difficulty of using `-1` as its counter intuitive to what you'd expect.

rlopez133:
Looking further into this, I'm thinking this should just be a documentation change. Things I'm running into 
* Ansible not respecting multiple when conditionals
* We get different behavior if its a hostname versus an IP address. For example, an IP address can be pingable, but if you then attempt https://<apivip>:6443 it shows -1 and no route to host. 
* No consistent error message, at times its Forbidden, Connection Refused

e-minguez:
> Looking further into this, I'm thinking this should just be a documentation change. Things I'm running into
> 
>     * Ansible not respecting multiple when conditionals
> 
>     * We get different behavior if its a hostname versus an IP address. For example, an IP address can be pingable, but if you then attempt https://:6443 it shows -1 and no route to host.
> 
>     * No consistent error message, at times its Forbidden, Connection Refused

I'm ok on adding a note in the docs instead.

DirectedSoul1:
> > Looking further into this, I'm thinking this should just be a documentation change. Things I'm running into
> > ```
> > * Ansible not respecting multiple when conditionals
> > 
> > * We get different behavior if its a hostname versus an IP address. For example, an IP address can be pingable, but if you then attempt https://:6443 it shows -1 and no route to host.
> > 
> > * No consistent error message, at times its Forbidden, Connection Refused
> > ```
> 
> I'm ok on adding a note in the docs instead.

Thanks @e-minguez  we will add this likely in Troubleshooting doc.

PR #400, opened 2020-05-22
Only poweroff if poweroff_servers has a host in it

johnbieren:
Signed-off-by: Johnny Bieren <jbieren@redhat.com>

# Description

Only call the task to power off all servers in the poweroff_servers group if at least one host was added to it. Otherwise, you get an error because the group won't actually exist.

## Type of change

## Testing

- [ ] Tested locally

**Test Configuration**:

- Versions: RHEL-8.1
- Hardware: Virtual

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
@@ -19,5 +19,6 @@
         password: "{{ hostvars[item]['ipmi_password'] }}"
         port: "{{ hostvars[item]['ipmi_port'] | default(623) }}"
         state: off
+      when: groups['poweroff_hosts'] is defined
       with_items: "{{ groups['poweroff_hosts'] }}"
   tags: powerservers


johnbieren:
[test-ipi]

ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
```diff
@@ -19,5 +19,6 @@
         password: "{{ hostvars[item]['ipmi_password'] }}"
         port: "{{ hostvars[item]['ipmi_port'] | default(623) }}"
         state: off
+      when: groups['poweroff_hosts'] is defined
```
rlopez133, line 22: I'm not sure we need this because lines 9,10 take care of what hosts should be part of the poweroff_hosts group. The masters should realistically always been in that group. 

What scenario are you running that gave you an error?

johnbieren: If all of your masters have poweroff=false, then the poweroff_hosts group never gets created and this task fails

rlopez133:
Even though what you are saying is true, there is not a scenario where we don't want the poweroff_hosts group to be undefined. This feature was included to add a little more flexibility for very specific use cases, but in general 95% of the time we want the playbook to go off and shutdown the systems (default behavior). If we include this when statement I see further issues in installing that are harder to debug due to this check passing.

PR #407, opened 2020-05-27
Allow provisioning network adjustments

### LINKED ISSUE #406: Missing some additional install-config options
The chapter in our doc that references additional install-config.yaml (https://openshift-kni.github.io/baremetal-deploy/4.4/Deployment.html#additional-install-config-parameters_ipi-install-prerequisites) options doesn't capture everything listed here: https://github.com/openshift/installer/blob/master/docs/user/metal/customization_ipi.md 

Specifically it was referenced we don't have provisioningDHCPRange

### LINKED ISSUE #410: [RFE] ability to override the provisioning network DHCP range and bootstrap VM IP address
Add the ability to override the provisioning network DHCP range and bootstrap VM IP address

dustinblack:
# Description

Add options to override the provisioning network DHCP range and bootstrap VM IP address. 

Fixes #410 

## Type of change

## Testing

Not yet tested.

## Checklist

- [ ] If a change is adding a feature, it should require a change to the README.md and the review should catch this. _**Docs have been relocated from README.md. Do I need to make a docs fix elsewhere?**_

### DIFF
--- a/ansible-ipi-install/inventory/hosts.sample
+++ b/ansible-ipi-install/inventory/hosts.sample
@@ -45,6 +45,10 @@ build=""
 # Provisioning IP address (default value)
 prov_ip=172.22.0.3
 
+# (Optional) Provisioning network configuration overrides (4.4+)
+#prov_bootstrap_ip=172.22.0.2
+#prov_dhcp_range="172.22.0.10,172.22.0.100"
+
 # (Optional) Enable playbook to pre-download RHCOS images prior to cluster deployment and use them as a local
 # cache. Default is false.
 #cache_enabled=True
--- a/ansible-ipi-install/roles/installer/templates/install-config.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config.j2
@@ -38,6 +38,15 @@ platform:
 {% if (release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int > 3)) %}
     provisioningNetworkInterface: {{ masters_prov_nic }}
     provisioningNetworkCIDR: {{ provisioning_subnet }}
+{% if prov_dhcp_range is defined and prov_dhcp_range|length %}
+    provisioningDHCPRange: {{ prov_dhcp_range }}
+{% endif %}
+{% if prov_bootstrap_ip is defined and prov_bootstrap_ip|length %}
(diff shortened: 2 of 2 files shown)


ansible-ipi-install/inventory/hosts.sample
```diff
@@ -45,6 +45,10 @@ build=""
 # Provisioning IP address (default value)
 prov_ip=172.22.0.3
 
+# (Optional) Provisioning network configuration overrides (4.4+)
+#prov_bootstrap_ip=172.22.0.2
+#prov_dhcp_range="172.22.0.10,172.22.0.100"
+
```
rlopez133, line 51: I don't think its this simple because we need to reflect that this could be an IPv6 bootstrap IP and range. I'll think about this more tomorrow

ansible-ipi-install/roles/installer/templates/install-config.j2
```diff
@@ -38,6 +38,15 @@ platform:
...
     provisioningNetworkCIDR: {{ provisioning_subnet }}
+{% if prov_dhcp_range is defined and prov_dhcp_range|length %}
+    provisioningDHCPRange: {{ prov_dhcp_range }}
+{% endif %}
+{% if prov_bootstrap_ip is defined and prov_bootstrap_ip|length %}
+    bootstrapProvisioningIP: {{ prov_bootstrap_ip }}
+{% endif %}
+{% if prov_ip is defined and prov_ip|length %}
+    clusterProvisioningIP: {{ prov_ip }}
+{% endif %}
```
rlopez133, line 49: you mention in the sample, 4.4+ maybe we need to add that logic in here as well to reflect that?

ansible-ipi-install/inventory/hosts.sample
dustinblack: I would think that as long as the `install-config.yml` is already able to accept IPv4 or IPv6, these values should be user-definable here, but I could be overlooking something for sure.

rlopez133: They do, I just mean that I want to think about what the example above should look like to better reflect that it could be either or.

ansible-ipi-install/roles/installer/templates/install-config.j2
rlopez133: You can likely piggyback off line 38.

PR #413, opened 2020-05-28
Lower down manifest number

### LINKED ISSUE #404: Rename 99_ manifests to 98_ as workaround for MCO bug
There's an ordering issue in MCO when using any custom manifests with
the 99_ prefix. Custom manifest until this is fixed should all start
with 98_ (or some lower number).

Bug here: https://github.com/openshift-metal3/dev-scripts/pull/1039

We need to fix our script to not create with 99_ prefix but use 98_ instead.

iranzo:
# Description

Rename manifests with 99- or 99_ to 98 to avoid bug https://github.com/openshift-metal3/dev-scripts/pull/1039

Fixes # (issue)

Closes #404

## Type of change

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
+++ b/ansible-ipi-install/roles/installer/tasks/40_create_manifest.yml
@@ -20,7 +20,7 @@
 - name: Copy the metal3-config.yaml to {{ dir }}/openshift directory
   copy:
     src: "{{ dir }}/metal3-config.yaml"
-    dest: "{{ dir }}/openshift/99_metal3-config.yaml"
+    dest: "{{ dir }}/openshift/98_metal3-config.yaml"
     owner: "{{ ansible_user }}"
     group: "{{ ansible_user }}"
     mode: '0644'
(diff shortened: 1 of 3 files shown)


iranzo:
[test-ipi]

iranzo:
Deploying on KNI3 succeeds it fails with ipi job :/

rlopez133:
@iranzo regarding 99-etc-chrony.conf.j2, since we are changing this, how about we change the template name to chrony.conf.j2 so that its not dependent on a number and leave the 98 pieces for the tasks?

iranzo:
> @iranzo regarding 99-etc-chrony.conf.j2, since we are changing this, how about we change the template name to chrony.conf.j2 so that its not dependent on a number and leave the 98 pieces for the tasks?

As we need both files (one is the actual conf and the other is the yaml), I've removed the XX-etc-chronny to leave etc-chrony and get the final number when it's rendered into final destination

rlopez133:
CI has been flaky, testing on my KNI4, will merge if success.

PR #478, opened 2020-07-30
Intelligently decide between using IPMI or RedFish protocol based on Vendor and Firmware version

### LINKED ISSUE #134: [RFE] for ansible-ipi-install: Ability to use redfish instead of ipmi
Currently the install-config.yaml only uses ipmi. We want the ability to use redfish or ipmi.

rlopez133:
# Description

Currently the Ansible playbook only allows for the use of IPMI protocol. This PR now contains logic to successfully decide if a particular host should use IPMI or RedFish as its protocol.

The logic is as follows:
* Dell server with iDRAC firmware version 4.20.20.20 or higher use RedFish 
* HPE server use RedFish 
* All other systems, or Dell systems that do not have a firmware of 4.20.20.20 or higher default to IPMI.

This patch will look at each server individually and make the appropriate decision thus removing the need for a user to know if their particular server and firmware can support RedFish. If it can, it will.  

In addition, there is now a `bmc_certificate_verification` variable that allows you to set to True if your RedFish certificates are self-signed. The default value is False. 

Fixes #134 

## Type of change

## Checklist

### DIFF
--- a/ansible-ipi-install/inventory/hosts.sample
+++ b/ansible-ipi-install/inventory/hosts.sample
@@ -132,6 +132,10 @@ pullsecret=""
 #root_device_hint="deviceName"
 #root_hint_value="/dev/sda"
 
+# (Optional) Disable BMC Certification Validation. When using self-signed certificates for your BMC, ensure to set to True.
+# Default value is False. 
+#disable_bmc_certificate_verification=True
+
 # Master nodes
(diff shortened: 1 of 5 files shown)


rlopez133:
[test-ipi]

iranzo:
[test-ipi]

PR #531, opened 2020-10-08
Adding bootMode option to install config

mukrishn:
# Description

This pull request is to let user configure an option introduced in 4.6 installer in install-config.yaml to choose boot mode per host. 
New option available - https://github.com/openshift/installer/blob/release-4.6/docs/user/metal/install_ipi.md#install-config

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/ansible-ipi-install/inventory/hosts.sample
+++ b/ansible-ipi-install/inventory/hosts.sample
@@ -141,6 +141,9 @@ pullsecret=""
 #provisioningHostIP=<baremetal_net_IP1>
 #bootstrapProvisioningIP=<baremetal_net_IP2>
 
+# (Optional) Change the boot mode of the OpenShift cluster nodes to legacy mode (BIOS). Default is UEFI. 
+#bootmode=legacy
+
 # Master nodes
 # The hardware_profile is used by the baremetal operator to match the hardware discovered on the host
 # See https://github.com/metal3-io/baremetal-operator/blob/master/docs/api.md#baremetalhost-status
--- a/ansible-ipi-install/roles/installer/templates/install-config.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config.j2
@@ -72,6 +72,9 @@ platform:
 {% endif %}
 {% endif %}
 {% if ((release_version[0]|int == 4) and (release_version[2]|int >= 6)) %}
+{% if bootmode is defined and bootmode == 'legacy' %}
+        bootMode: legacy
+{% endif %}
 {% if 'root_device_hint' in hostvars[host] and 'root_device_hint_value' in hostvars[host] and hostvars[host]['root_device_hint'] in roothint_list %}
         rootDeviceHints:
           {{ hostvars[host]['root_device_hint'] }}: {{ hostvars[host]['root_device_hint_value'] }}
@@ -103,6 +106,9 @@ platform:
 {% endif %}
 {% endif %}
 {% if ((release_version[0]|int == 4) and (release_version[2]|int >= 6)) %}
+{% if bootmode is defined and bootmode == 'legacy' %}
+        bootMode: legacy
+{% endif %}
 {% if 'root_device_hint' in hostvars[host] and 'root_device_hint_value' in hostvars[host] %}
         rootDeviceHints:
           {{ hostvars[host]['root_device_hint'] }}: {{ hostvars[host]['root_device_hint_value'] }}
--- a/ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
@@ -74,6 +74,9 @@ platform:
 {% endif %}
 {% endif %}
 {% if ((release_version[0]|int == 4) and (release_version[2]|int >= 6)) %}
+{% if bootmode is defined and bootmode == 'legacy' %}
+        bootMode: legacy
+{% endif %}
 {% if 'root_device_hint' in hostvars[host] and 'root_device_hint_value' in hostvars[host] and hostvars[host]['root_device_hint'] in roothint_list %}
         rootDeviceHints:
           {{ hostvars[host]['root_device_hint'] }}: {{ hostvars[host]['root_device_hint_value'] }}
@@ -105,6 +108,9 @@ platform:
 {% endif %}
 {% endif %}
 {% if ((release_version[0]|int == 4) and (release_version[2]|int >= 6)) %}
+{% if bootmode is defined and bootmode == 'legacy' %}
+        bootMode: legacy
+{% endif %}
 {% if 'root_device_hint' in hostvars[host] and 'root_device_hint_value' in hostvars[host] %}
         rootDeviceHints:
           {{ hostvars[host]['root_device_hint'] }}: {{ hostvars[host]['root_device_hint_value'] }}


smalleni:
@rlopez133 PTAL :)

rlopez133:
Hey @mukrishn thanks for the PR! 

I was thinking a better way to handle this so we don't include it as another host var but instead make it a variable within the inventory file itself. 

For example, within hosts.sample having something at the end of the section labeled
```
######################################
# Vars regarding install-config.yaml #
######################################
```

Something like 

```
# (Optional) Change the boot mode of the OpenShift cluster nodes to legacy mode (BIOS). Default is UEFI.
#bootmode=legacy
```

Then within the roles/node-prep/defaults/main.yml
```
bootmode: UEFI
```
Then within 
roles/installer/templates/install-config.j2

 under line 66 and 97
```
{% if bootmode == 'legacy' %}
      bootmode: legacy
{% endif %}
```
Then within 
roles/installer/templates/install-config-virtualmedia.j2

 under line 68 and 99
```
{% if bootmode == 'legacy' %}
      bootmode: legacy
{% endif %}
```

smalleni:
> Hey @mukrishn thanks for the PR!
> 
> I was thinking a better way to handle this so we don't include it as another host var but instead make it a variable within the inventory file itself.
> 
> For example, within hosts.sample having something at the end of the section labeled
> 
> ```
> ######################################
> # Vars regarding install-config.yaml #
> ######################################
> ```
> 
> Something like
> 
> ```
> # (Optional) Change the boot mode of the OpenShift cluster nodes to legacy mode (BIOS). Default is UEFI.
> #bootmode=legacy
> ```
> 
> Then within the roles/node-prep/defaults/main.yml
> 
> ```
> bootmode: UEFI
> ```
> 
> Then within
> roles/installer/templates/install-config.j2
> 
> under line 66 and 97
> 
> ```
> {% if bootmode == 'legacy' %}
>       bootmode: legacy
> {% endif %}
> ```
> 
> Then within
> roles/installer/templates/install-config-virtualmedia.j2
> 
> under line 68 and 99
> 
> ```
> {% if bootmode == 'legacy' %}
>       bootmode: legacy
> {% endif %}
> ```
> 
> Forgot to add that that if statement will likely require condition to work with 4.6+

@rlopez133 Wouldn't we lose the flexibility of this variable to be a per host variable if we do that?

DirectedSoul1:
@smalleni  that's a good point, but adding the bootMode option per host might be tricky 

-  So, each master node in `UEFI` and workers in `Legacy BIOS` might increase latency and bring up timing issues
- also, debugging, in general, might be difficult 

so, @rlopez133 's suggestions make it easier for deployments, we are open to further suggestions. Thanks

rlopez133:
> @smalleni that's a good point, but adding the bootMode option per host might be tricky
> 
> * So, each master node in `UEFI` and workers in `Legacy BIOS` might increase latency and bring up timing issues
> * also, debugging, in general, might be difficult
> 
> so, @rlopez133 's suggestions make it easier for deployments, we are open to further suggestions. Thanks

100% agree here with @DirectedSoul1 . I don't think we want flexibility for each host to change it. I think you should pick a standard of your boot mode for all your hosts. I think this will likely give better results/chances to succeed in an install.

mukrishn:
@rlopez133 @DirectedSoul1 I agree with your concerns, but this config help us a lot to build a heterogeneous node cluster. We are targeting a 500 node deployment and we have different classes and models of servers in our lab.

DirectedSoul1:
@mukrishn  totally legit use case, is the plan to categorize control plane nodes in one `bootMode` and worker nodes in other mode or to mix `bootMode` even for control plane nodes?

smalleni:
Just had a chat with Roger on slack and I see his cconerns about  making things harder for debug with Mixed bootmode. Let us try and get the bootmode enforced to all hosts in the cluster with this patch and once it is well tested and we feel fairly confident, we can push for a bootmode per host....

@mukrishn

mukrishn:
@rlopez133 Okay Roger, totally agreeing the difficulties that it brings. I will update as per your suggestion.

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -200,3 +200,6 @@ provisioner.example.com
 #   content sources for the local registry. The contents of this file
 #   will be appended to the install-config.yml file.
 # disconnected_registry_mirrors_file=/path/to/install-config-appends.json
+
+# (Optional) Change the boot mode of the OpenShift cluster nodes to legacy mode (BIOS). Default is UEFI. 
+#bootmode=legacy
```
rlopez133, line 205: Apologies because I'm being a nit here, but can we add this after line 142? This way this is part of the "Vars regarding install-config.yaml"

ansible-ipi-install/roles/installer/templates/install-config.j2
```diff
@@ -106,9 +106,9 @@ platform:
 {% endif %}
 {% endif %}
 {% if ((release_version[0]|int == 4) and (release_version[2]|int >= 6)) %}
-{% if 'boot_mode' in hostvars[host] %}
-        bootMode: {{ hostvars[host]['boot_mode'] }}
-{% endif %}
+{% if bootmode == 'legacy' %}
+        bootMode: legacy
+{% endif %}
```
rlopez133, line 111: Can you add the same piece of code for the install-config-virtualmedia.j2 file as well?

ansible-ipi-install/roles/installer/defaults/main.yml
```diff
@@ -14,3 +14,4 @@ webserver_caching_port_container: 8080
 registry_creation: false
 url_passed: false
 httpd_cache_files: "{{ provision_cache_store }}httpd/"
+bootmode: UEFI
```
rlopez133, line 17: complaining that there is no newline at end of file.

rlopez133 (changes requested):

ansible-ipi-install/roles/installer/templates/install-config.j2
mukrishn: @rlopez133 I couldn't find that file in that path so I ignored ? could you point me to the file please.

ansible-ipi-install/inventory/hosts.sample
mukrishn: No worries

ansible-ipi-install/roles/installer/templates/install-config.j2
DirectedSoul1: @mukrishn are you looking for the below file?
 https://github.com/openshift-kni/baremetal-deploy/blob/master/ansible-ipi-install/roles/installer/templates/install-config.j2 , this should be the path of that template file.

mukrishn: @DirectedSoul1 I am looking for this template file - `install-config-virtualmedia.j2`

rlopez133: I'm guessing you don't see it because @smalleni hasn't rebased with the newest changes? I added this very recently. Its in the same templates dir

mukrishn:
@rlopez133 @smalleni I have incorporated all the changes and I tested the script w/ and w/o bootmode.

smalleni:
@rlopez133 I think for squash and merge it coalesces all the commit messages into one. Let me know if you want to try that or we can ask @mukrishn to squash.

ansible-ipi-install/roles/installer/defaults/main.yml
rlopez133, line 17: I realized we don't need this at all. There is no check for it.

smalleni: Wouldn't we need because if bootmode is not set in the inventory (it's commented out by default), our templating of install-config.yml will fail. If we want to remove this default variable, maybe then we need to add an additional check in the install-config.j2 to check if bootmode is defined 
```
{% if bootmode is defined and bootmode == 'legacy' %}
        bootMode: legacy
{% endif %}

File: ansible-ipi-install/inventory/hosts.sample [hunk 4]
```diff
@@ -141,6 +141,9 @@ pullsecret=""
 #provisioningHostIP=<baremetal_net_IP1>
 #bootstrapProvisioningIP=<baremetal_net_IP2>
 
+# (Optional) Change the boot mode of the OpenShift cluster nodes to legacy mode (BIOS). Default is UEFI.
```
smalleni, line 144: Can you add a comment saying, if this is commented or set to anything other than legacy, UEFI is used.

ansible-ipi-install/roles/installer/defaults/main.yml
mukrishn: @smalleni okay i will add a `if defined` check and remove that default variable.

ansible-ipi-install/inventory/hosts.sample
```diff
@@ -141,6 +141,10 @@ pullsecret=""
 #provisioningHostIP=<baremetal_net_IP1>
 #bootstrapProvisioningIP=<baremetal_net_IP2>
 
+# (Optional) Change the boot mode of the OpenShift cluster nodes to legacy mode (BIOS). Default is UEFI. 
+# If this is commented or set to anything other than "legacy", UEFI will be used.
```
rlopez133, line 145: can we remove this line then its redundant to the one above it.

rlopez133:
@mukrishn would you mind squashing all the commit messges to just one? I'll merge this in right after.

mukrishn:
@rlopez133, Now all in one commit.

PR #557, opened 2020-11-03
fix(100_power_off_cluster_servers): skips errors on the power off tas…

### LINKED ISSUE #555: "'dict object' has no attribute 'poweroff_hosts'"
The playbook fails with the following error when all the inventory hosts have the poweroff parameter set to False:

TASK [node-prep : Power off hosts] ***********************************************************************************************************************************************************************************************************
Friday 30 October 2020  17:31:31 +0000 (0:00:00.141)       0:02:32.848 ********                                                                                                                                                               
fatal: [---]: FAILED! => {"msg": "'dict object' has no attribute 'poweroff_hosts'"}

iranzo:
As we prepare the list, it might be empty, this avoids errors to halt script

Closes #555

# Description

Fixes # (issue)

## Type of change

## Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Versions:
- Hardware:

## Checklist

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
@@ -24,4 +24,5 @@
       retries: 10
       delay: 5
       with_items: "{{ groups['poweroff_hosts'] }}"
+      when: groups['poweroff_hosts'] is defined
   tags: powerservers


iranzo:
[test-ipi]

ansible-ipi-install/roles/node-prep/tasks/100_power_off_cluster_servers.yml
```diff
@@ -24,4 +24,5 @@
       retries: 10
       delay: 5
       with_items: "{{ groups['poweroff_hosts'] }}"
+      ignore_errors: true
```
rlopez133, line 27: we don't want to ignore because if we ignore, its possible it will mess up the entire install. It's better it fail where it fails.

iranzo: Should then include some clause of it not being empty? because actual status is that list is empty and instead of ignoring it, it tries to use it and fails (groups['poweroff_hosts'])

In python when you do:

~~~py
var=[1,2,3]

for item in var:
    print(item)

# or
var = []
for item in var:
    print(item)

~~~

In both cases, it works, as first it loops over three elements and second it loops over 0 elements.

In this case it reacts more like a dictionary var with undefined element and it fails completely

iranzo: I've updated the PR to check if the group is defined and if the number of elements is higher than 0, to simulate above Python behavior

rlopez133: Yes. I think checking if the group is defined is the way to go /lgtm

Issue #568, opened 2020-11-05
Within additional install parameters required section no description of sshKey or pullsecret


rlopez133:
https://openshift-kni.github.io/baremetal-deploy/4.6/Deployment.html#additional-install-config-parameters_ipi-install-configuration-files

We need to add a description for sshKey and pullsecret as part of the required parameters table.

PR #575, opened 2020-11-10
Added sshKey and pullSecret descriptions.

johnwilkins:
Fixes: #568

@rlopez133 

# Description

Added sshKey and pullSecret descriptions.


rlopez133:
This needs more details.

For sshKey something like:

The sshKey is the id_rsa.pub of the system you wish to use to access your control plane nodes and worker nodes from. Typically it is the pub key of your provision host. Without providing a key, you could not log into your control plane nodes or worker nodes.

For pullsecret something like:

A copy of the pull secret that was retrieved from https://cloud.redhat.com/openshift/install/metal/user-provisioned

PR #581, opened 2020-11-13
Updating the upstream doc to reflect "installer-provisioned"

johnwilkins:
Updating the upstream doc to reflect "installer-provisioned" rather than IPI. Changed worker node description too. Removed the configuring networks topic as it was just a link to an unpublished KCS article. Updated callouts in install-config.yaml. Also restored conditional text that was removed in the downstream doc.


iranzo (approved):
/lgtm

PR #586, opened 2020-11-16
Putting periods outside of conditional tags so punctuation is correct.

johnwilkins:

# Description
@rlopez133 
@iranzo 

Putting periods outside of conditional tags so punctuation is correct.

**Question**: Are the ifdef::upstream[] conditionals still relevant in this doc? They seem to apply only to IPv6 examples. Can they be removed?

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
+++ b/documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
@@ -3,7 +3,6 @@
 // * installing/installing_bare_metal_ipi/ipi-install-installation-workflow.adoc
 
 [id="ipi-install-bmc-addressing_{context}"]
-[[bmcaddressing]]
 
 = BMC addressing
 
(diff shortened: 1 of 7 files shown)


documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
```diff
@@ -3,7 +3,6 @@
 // * installing/installing_bare_metal_ipi/ipi-install-installation-workflow.adoc
 
 [id="ipi-install-bmc-addressing_{context}"]
-[[bmcaddressing]]
```
iranzo, line 6: This reference is also used in install-config-yaml for each release, if we remove it, we'll need to update all the other ones

iranzo (changes requested):
Thanks a lot John!, can you check the one on bmcaddressing?

documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
iranzo: I've made a test by renaming them on my own, but the _context is not expanded when used in the install-config-yaml$RELEASE so I would leave this anchor here and work in another PR on renaming/removing so that it still works

iranzo:
In regards to the ifupstream, I think those can be removed here, it's something we added first for the 'DEVEL' version of the document but now that ipv6 has landed, the version check should be enough.

documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
iranzo: Raised https://github.com/openshift-kni/baremetal-deploy/issues/589 on this topic

iranzo: Solved with the new commit I made

iranzo (approved):
/lgtm
/approve

PR #587, opened 2020-11-16
Added day 2 operation for adding a bare metal node.

johnwilkins:
@rlopez133 
@iranzo 
# Description

Took draft from Roger and re-worked it a bit. The main changes are: 

I've added the steps: 
1. **Get a count of the worker nodes.** This is because if they already installed n number of worker nodes, and we don't know how many they have, we need that number to scale the nodes say from 2 to 3, because it appears that the replicas is the hard count, it's not incrementing them by the count.
2. **Get the machine set.** We don't know what the name of the machine set is. So I added that step so that it's obvious. 

I've consolidated the step: 

- **The baremetal node changes to provisioning status.** step with **When the node is provisioned it moves to provisioned status.** The command line usage is identical. What we're talking about is monitoring the state transition. 

I've also broken multiple commands in one block into multiple blocks, and put outputs from commands into their own blocks. I used the same corresponding steps that I had in my draft for cleaner punctuation and style compliance. 

Fixes https://issues.redhat.com/browse/OSDOCS-1308

I will have to create a separate downstream version with a separate assembly.

## Type of change

## Checklist

### DIFF
--- a/documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
+++ b/documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
@@ -14,12 +14,12 @@
 +
 [source,bash]
 ----
-[kni@openshift-master-0 ~]$ echo -ne "root" | base64
+[kni@provisioner ~]$ echo -ne "root" | base64
 ----
 +
 [source,bash]
 ----
-[kni@openshift-master-0 ~]$ echo -ne "calvin" | base64
+[kni@provisioner ~]$ echo -ne "calvin" | base64
 ----
 
 
@@ -27,7 +27,7 @@
 +
 [source,bash]
 ----
-[kni@openshift-master-0 ~]$ vim bmn.yaml
+[kni@provisioner ~]$ vim bmh.yaml
 ----
 +
 [source,yaml]
@@ -59,14 +59,14 @@ Replace `<num>` for the worker number of bare metal node in two `name` fields an
 [NOTE]
 ====
 When using `redfish` or `redfish-virtualmedia`, add the
-appropriate addressing as described in the BMC addressing section. See xref:ipi-install-bmc-addressing_{context}[BMS addressing] for details.
+appropriate addressing as described in the BMC addressing section. See xref:ipi-install-bmc-addressing_{context}[BMC addressing] for details.
 ====
 
 . Create the bare metal node.
 +
 [source,bash]
 ----
-[kni@openshift-master-0 ~]$ oc -n openshift-machine-api create -f bmn.yaml
+[kni@provisioner ~]$ oc -n openshift-machine-api create -f bmh.yaml
 ----
 +
 [source,bash]
@@ -81,7 +81,7 @@ Where `<num>` will be the worker number.
 +
 [source,bash]
 ----
-[kni@openshift-master-0 ~]$ oc -n openshift-machine-api get bmn openshift-worker-<num>
+[kni@provisioner ~]$ oc -n openshift-machine-api get bmh openshift-worker-<num>
 ----
 +
 Where `<num>` is the worker node number.
@@ -96,7 +96,7 @@ openshift-worker-<num>   OK       ready                            ipmi://<out-o
 +
 [source,bash]
 ----
-[kni@openshift-master-0 ~]$  oc -n openshift-machine-api get bmn openshift-worker-<num>
+[kni@provisioner ~]$  oc -n openshift-machine-api get bmh openshift-worker-<num>
 ----
 +
 Where `<num>` is the worker node number.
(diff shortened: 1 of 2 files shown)


documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
```diff
@@ -0,0 +1,182 @@
...
+    address: ipmi://<bmc-ip>
+    credentialsName: openshift-worker-<num>-bmc-secret
+----
++
+Replace `<num>` for the worker number of bare metal node in two `name` fields and `credentialsName` field. Replace `<base64-of-uid>` with the `base64` string of the username. Replace `<base64-of-pwd>` with the `base64` string of the password. Replace `<NIC1-mac-address>` with the MAC address of the bare metal node's first NIC. Replace `<bmc-ip>` with the IP address of the bare metal node's baseboard management controller.
+
+[NOTE]
+====
+When using `redfish` or `redfish-virtualmedia`, add the
+appropriate addressing as described in the BMC addressing section. See xref:ipi-install-bmc-addressing_{context}[BMS addressing] for details.
```
rlopez133, line 62: spelled wrong, should be BMC not BMS, but it doesn't properly link

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
```diff
@@ -0,0 +1,182 @@
...
+====
+
+. Create the bare metal node.
++
+[source,bash]
+----
+[kni@openshift-master-0 ~]$ oc -n openshift-machine-api create -f bmn.yaml
```
rlopez133, line 69: lets change this to from bmn.yaml to bmh.yaml to be mroe aligned with how its called 'baremetal host'

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
```diff
@@ -0,0 +1,182 @@
...
+baremetalhost.metal3.io/openshift-worker-<num> created
+----
++
+Where `<num>` will be the worker number.
+
+. Power up and inspect the bare metal node.
++
+[source,bash]
+----
+[kni@openshift-master-0 ~]$ oc -n openshift-machine-api get bmn openshift-worker-<num>
```
rlopez133, line 84: command is `bmh` not bmn`

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
```diff
@@ -0,0 +1,182 @@
...
+----
+NAME                 STATUS   PROVISIONING STATUS   CONSUMER   BMC                 HARDWARE PROFILE   ONLINE   ERROR
+openshift-worker-<num>   OK       ready                            ipmi://<out-of-band-ip>   unknown            true
+----
+
+. Ensure the `PROVISIONING STATUS` is `ready` before provisioning the bare metal node.
++
+[source,bash]
+----
+[kni@openshift-master-0 ~]$  oc -n openshift-machine-api get bmn openshift-worker-<num>
```
rlopez133, line 99: command is `bmh` not `bmn`

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
rlopez133: everywhere we say openshift-master-0, I think we should label this provisioner. If we decide not to go that route and use master-0 then we need to export the KUBECONFIG to the master host as those commands won't work without that.

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
```diff
@@ -0,0 +1,182 @@
...
+[source,bash]
+----
+[kni@openshift-master-0 ~]$ oc get nodes
+----
+
+. Get the machine set.
++
+[source,bash]
+----
+[kni@openshift-master-0 ~]$ oc get machinesets -n openshift-machine-api
```
rlopez133, line 121: we should show output.

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
```diff
@@ -0,0 +1,182 @@
...
+You can also check the kubelet.
++
+[source,bash]
+----
+[kni@openshift-master-0 ~]$ ssh openshift-worker-<num>
+----
++
+[source,bash]
+----
+[kni@openshift-worker-<num>]$ journalctl -fu kubelet
```
rlopez133, line 181: after this, comes 13.3.1 which is in the wrong spot. 

Take a look https://deploy-preview-587--baremetal-deploy.netlify.app/4.6/deployment#ipi-install-upstream-dns-records-for-the-worker

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
johnwilkins: We need to be using the [id="anchor_{context}"] approach were (context) is set in the master.adoc, for example: :context: some-context. We should not be using [[anchor-tag]].

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
johnwilkins: What is the right spot?

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
iranzo: It's been fixed with current 'master' version

documentation/ipi-install/modules/ipi-install-upstream-adding-a-worker-node-to-an-existing-cluster.adoc
iranzo: @rlopez133 (ping)

PR #595, opened 2020-11-23
Incorporating style changes from Ashley downstream.

johnwilkins:
@iranzo 
# Description

Ashley had some style changes. We use "second" instead of "2nd", "and so on" instead of "etc." Also, we still do not support xref tags in modules. It's valid asciidoc, but it will get us push back on pull requests.


iranzo (approved):
/lgtm

PR #605, opened 2020-12-07
Added references to the Day 0 firmware requirements for virtual media.

johnwilkins:

@rlopez133 
@iranzo 

# Description

In lieu of a Day 1 operation, the BMC addressing section contains a note referring virtual media users to the firmware requirements section. 

Fixes # Telcodocs 109

## Type of change

### DIFF
--- a/documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
+++ b/documentation/ipi-install/modules/ipi-install-bmc-addressing.adoc
@@ -118,6 +118,24 @@ platform:
           password: <password>
 ----
 
+While it is recommended to have a certificate of authority for the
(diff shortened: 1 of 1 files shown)


omichaeli:
@johnwilkins I see that the `disableCertificateVerification: True` example for HPE and Dell virtual media is missing.
Is that intentional?

rlopez133:
> @johnwilkins I see that the `disableCertificateVerification: True` example for HPE and Dell virtual media is missing.
> Is that intentional?

Good catch. It is missing. It should follow the same layout as we show for RedFish for HPE and RedFish for Dell, where we give the example without and then point out if you can disable it.

PR #614, opened 2020-12-15
Initial cut of iRMC for Fujitsu. 

johnwilkins:
@rlopez133 
@iranzo 

# Description

This is an initial cut based on https://issues.redhat.com/browse/TELCODOCS-111. I based this on the content received in telcodocs 111 and conversation with Dmitri. However, it seems to me that we should also include sections for RedFish and RedFish with Virtual Media for Fujitsu, since we only support SecureBoot with RedFish Virtual Media. According to their S5 spec, they do have RedFish and Virtual media. See https://www.reckzigel.at/wp-content/uploads/2018/03/irmc-s5-configuration-en.pdf for additional details.

Fixes # telcodocs-111

## Type of change


iranzo:
Hi @johnwilkins if it only works for iRMC S5 or higher,  should we add a note about it? Will it work with all OCP versions or only in 4.6+ ? (to add a conditional if that's the case)

PR #628, opened 2021-01-11
Upstream downstream sync on worklow

johnwilkins:
# Description

I'm not sure if the upstream build is using Deployment.adoc or ipi-install-upstream-master.adoc, but the build SHOULD use the assemblies rather than module-by-module inclusion. This PR contains a number of changes to help synchronize the upstream and downstream repos. I will need assistance testing it, since my local toolchain for some reason really doesn't like the ifeval statements, but I do not see any obvious problems. 

@iranzo 
@rlopez133 

## Type of change

## Checklist


iranzo:
image


Deployment.adoc is the same as ipi-install-upstream-master.adoc

iranzo:
About the verification, whatever is rendered with netlify for this PR (https://deploy-preview-628--baremetal-deploy.netlify.app/) if it looks good, it should be valid, as the repo itself uses a very similar method by executing asciidoctor in the way as it's defined in `utils/build.sh`

iranzo (changes requested):
image

The changes break the document generation

iranzo:
@johnwilkins can you tell me how are you building the docs? I think it fails because you're building from a different folder than the one we use.

I don't mind adapting repo's build scripts to work in the more similar way to downstream, so please, can you provide some guidance on how the downstream ones are built and how are you building them?  

In my case, I always rely on netlify rendering unless I'm doing other major changes to the site which is more 'jekyll' related than the docs

Thanks

johnwilkins:
Downstream, they have a topic.yaml file in the top-level directory that references the assemblies directly. That's what's building them. I think you had a soft link I didn't pick up when I was editing, so I probably have the references incorrect in the upstream-master.adoc file.

rlopez133:
speaking to @johnwilkins seems all is building now with this PR. @iranzo is it safe to merge?

iranzo:
image


@rlopez133 @johnwilkins The menus seems to look nice but the inclusion of the attributes seems to look awkward on the generated file. (https://deploy-preview-628--baremetal-deploy.netlify.app/4.6/deployment) 

Seems like the vars are not replaced properly:

image

Any idea on that?

iranzo:
Additionally, I was able to build something locally via setting the values for the first two vars included in common-attributes 

{product-author}
{product-version}

With a change in build script to define them, on netlify still complains and gets the other vars showing the 'code' but not the value, but at least locally with asciidoctor it worked.

It also complained of missing kebap.png file in the repo, and when using "icons: " we're not using the 'font' as we use in the other modules, so extra images might be missing in the document generation, while, using 'icons: font' it does use a text font to include them.

Any idea on the first part @johnwilkins ?

PR #651, opened 2021-02-04
Add vmedia support for workers

derekhiggins:
Masters were being deployed with vmedia while workers
we're not.

### DIFF
--- a/ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
@@ -88,14 +88,18 @@ platform:
       - name: {{ hostvars[host]['inventory_hostname'] }}
         role: {{ hostvars[host]['role'] }}
         bmc:
-{% if groups['dell_hosts_redfish'] is defined and host in groups['dell_hosts_redfish'] %}
-          address: redfish://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/System.Embedded.1
+{% if groups['dell_hosts_redfish'] is defined and host in groups['dell_hosts_redfish'] and ((release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int >= 6))) and enable_virtualmedia|bool %}
+          address: idrac-virtualmedia://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/System.Embedded.1
           disableCertificateVerification: {{ disable_bmc_certificate_verification }}
-{% elif groups['hp_hosts_redfish'] is defined and host in groups['hp_hosts_redfish'] %}
-          address: redfish://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/1
+{% elif groups['hp_hosts_redfish'] is defined and host in groups['hp_hosts_redfish'] and ((release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int >= 6))) and enable_virtualmedia|bool %}
+          address: redfish-virtualmedia://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/1
+          disableCertificateVerification: {{ disable_bmc_certificate_verification }}
+{% elif ansible_system_vendor == 'Dell Inc.' %}
+          address: idrac-virtualmedia://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/System.Embedded.1
           disableCertificateVerification: {{ disable_bmc_certificate_verification }}
 {% else %}
-          address: ipmi://{{ hostvars[host]['ipmi_address']|ipwrap }}:{{ hostvars[host]['ipmi_port']|default(623) }}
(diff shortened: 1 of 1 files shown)


iranzo (commented):
[test-ipi]

fredericlepied:
test

ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
```diff
@@ -88,14 +88,18 @@ platform:
...
         bmc:
-{% if groups['dell_hosts_redfish'] is defined and host in groups['dell_hosts_redfish'] %}
-          address: redfish://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/System.Embedded.1
+{% if groups['dell_hosts_redfish'] is defined and host in groups['dell_hosts_redfish'] and ((release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int >= 6))) and enable_virtualmedia|bool %}
+          address: idrac-virtualmedia://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/System.Embedded.1
           disableCertificateVerification: {{ disable_bmc_certificate_verification }}
-{% elif groups['hp_hosts_redfish'] is defined and host in groups['hp_hosts_redfish'] %}
-          address: redfish://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/1
+{% elif groups['hp_hosts_redfish'] is defined and host in groups['hp_hosts_redfish'] and ((release_version[0]|int > 4) or ((release_version[0]|int == 4) and (release_version[2]|int >= 6))) and enable_virtualmedia|bool %}
+          address: redfish-virtualmedia://{{ hostvars[host]['ipmi_address']|ipwrap }}/redfish/v1/Systems/1
```
iranzo, line 95: https://bugzilla.redhat.com/show_bug.cgi?id=1917751#c7  redfish:// should be ok according to BZ

derekhiggins: I've only copied the logic used for the master nodes higher up in the file, 

setting to "redfish://" wont use virtual media, in that bug Dmitry is saying that you can use redfish:// on Dell hosts but you can't use redfish-virtual-media:// , this is a hp host so should be using redfish-virtual-media (if deploying with vmedia).

rlopez133: Hey @derekhiggins  and @iranzo , 

Yes please use the same logic from masters for workers if you plan to update this. If you want to use virtual_media there is a flag you need to set in the inventory file to make that happen. 

redfish:// -- just redfish protocol (all vendors)
idrac-virtualmedia:// (Dell Virtual Media)
redfish-virtualmedia:// (HPE Virtual Media)

The goal of this is to attempt to use redfish:// when systems can use it. However, if user enables virtual_media flag in the inventory file it can use that instead. 

The logic here is forcing people to use virtual media and that should not be. It should default to redfish when it can and have user say they want virtual media in the inventory file. (I can't recall the flag name right now).

rlopez133: To add, if the code can't define redfish, it defaults to ipmi://

derekhiggins: > Hey @derekhiggins and @iranzo ,

Hi, I'm a little confused with your comments, see below

> 
> Yes please use the same logic from masters for workers if you plan to update this. If you want to use virtual_media there is a flag you need to set in the inventory file to make that happen.

I **am** using the same logic that is being used for the masters (40 lines above).
This template is only used when `enable_virtualmedia` is set.

> 
> redfish:// -- just redfish protocol (all vendors)
> idrac-virtualmedia:// (Dell Virtual Media)
> redfish-virtualmedia:// (HPE Virtual Media)
> 
> The goal of this is to attempt to use redfish:// when systems can use it. However, if user enables virtual_media flag in the inventory file it can use that instead.
> 
> The logic here is forcing people to use virtual media and that should not be. It should default to redfish when it can and have user say they want virtual media in the inventory file. (I can't recall the flag name right now).

Yes, this logic is forcing people to use virtual media, if they didn't want to use vmedia then the other template would be used (ansible-ipi-install/roles/installer/templates/install-config.j2)

iranzo: For me, it makes sense  to keep both sections, for masters and workers in the same way

iranzo:
Verified installation on cluster 1 OSD
image

ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
rlopez133: Youre right @derekhiggins -- did I mention we just had a baby? lol Obviously the lack of sleep can seen above lol. 

Continue on and ignore me lol. And thanks for the update!

derekhiggins: > Youre right @derekhiggins -- did I mention we just had a baby? lol Obviously the lack of sleep can seen above lol.
> 
> Continue on and ignore me lol. And thanks for the update!

been there, congratulations on the new arrival. ,

PR #681, opened 2021-03-02
Per https://bugzilla.redhat.com/show_bug.cgi?id=1923965, I modified BMC docs

johnwilkins:
# Description

The BZ calls for an address table with the vendors. I thought it might make more sense to segregate the vendors into different modules to reduce the scope for the reader, and also include the table at the beginning. I also re-ordered it so that the most favored approach--RedFish Virtual Media--is first and IPMI is last since it isn't secure. Will restore the Fujitsu docs in a separate PR.

Fixes: BZ-1923965

@iranzo

## Type of change


iranzo:
image
Shouldn't be the idrac + rootdevice hints one level lower?

PR #700, opened 2021-03-26
Second draft of NTP docs.

johnwilkins:

# Description

Created a second draft of NTP docs incorporating Pablo's comments and trying to make the text a bit clearer. Also found an uncommented release variable in an assembly (needed for downstream, but breaks things in upstream).

Fixes # TELCODOCS-115

Preview at: https://deploy-preview-700--baremetal-deploy.netlify.app/4.8/deployment#configuring-ntp_ipi-install-post-installation-configuration

## Type of change

### DIFF
--- a/documentation/ipi-install/ipi-install-post-installation-configuration.adoc
+++ b/documentation/ipi-install/ipi-install-post-installation-configuration.adoc
@@ -1,0 +1,8 @@
+[id="ipi-install-post-installation-configuration"]
+= Post-installation configuration
+include::modules/common-attributes.adoc[]
+:context: ipi-install-post-installation-configuration
+
+After successfully deploying an installer-provisioned cluster, consider the following post-install procedures.
+
+include::modules/ipi-install-configuring-ntp.adoc[leveloffset=+1]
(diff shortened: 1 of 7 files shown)


iranzo:
~~~
4.1. Configuring NTP
~~~
The OpenShift installation already configures the NTP client on the hosts (masters/workers) against the one defined in the install-config.yaml, this might is an extension to create an NTP server vs only the client.

Would it make sense to say Configuring a in-cluster NTP Server for disconnected operations or something along the lines?

~~~
OpenShift Container Platform 4.8 and later releases install NTP server on the control plane nodes. After successfully deploying an installer-provisioned cluster, configure the NTP servers on the control plane nodes, and configure worker nodes as NTP clients.
~~~
OpenShift CP XXX install NTP client on the cluster nodes, using this procedure an NTP server will be deployed on the control plane nodes and workers would be configured as clients. ?


~~~
OpenShift Container Platform nodes must agree on a date and time to run properly. When worker nodes retrieve the date and time from the NTP servers on the control plane nodes, it enables both the installation of clusters connected to a routable network with access to upstream NTP servers, or clusters that require additional security such that they are not connected to a routable network and thereby don’t have access to upstream NTP servers.
~~~

With the configuration we've defined in the PR, we're not reaching to external NTP servers, we can do that by specifying them via a pool pointing to upstream servers., so I would rephrase as:

OSCP must agree on date and time to run properly. When *the* nodes cannot retrieve date and time from upstream NTP servers , for example, with clusters not connected to a routable network, this procedure enables the workers to get the time from the NTP servers defined on the control plane nodes.

What do you think?

Thanks!
Pablo

PR #910, opened 2022-10-25
Add support to pass NMstate vars in install-config

### LINKED ISSUE #909: add support to include NMstate network configuration for nodes in install-config.yaml
Starting in 4.10 a new way to setup networking in the OCP nodes is available using NMstate parameters at day1[1]. As a consumer of this repository I would like to have an option to setup the nodes network configuration with the new method.

[1]
https://docs.openshift.com/container-platform/4.10/installing/installing_bare_metal_ipi/ipi-install-installation-workflow.html#configuring-host-network-interfaces-in-the-install-config-yaml-file_ipi-install-installation-workflow

manurodriguez:
This patch allows to include NMstate network settings in the install-config.yaml file and setup custom
networking at day1

# Description

The change allows to pass `networkConfig` settings in YAML for OCP nodes in the install-config.yaml file via two possible ways, either by loading configurations from specified files, or by defining the variables in yaml format.
This feature is available since OCP 4.10 and we want to have an additional option to setup networking besides creating machine-config manifests with NetworkManager connections.

Fixes #909 

## Type of change

- [ ] This change is a documentation update

## Testing

OCP Deployment Tests are performed in CI Labs, to validate:
- [x] The change does not introduce issues without the settings in both OCP >= 4.10 and below
- [x] The NMstate settings are passed to the install-config.yaml and clusters are deployed using custom network config for  OCP >= 4.10

**Test Configuration**:

- Versions: OCP 4.8, 4.9, 4.10, 4.11
- Hardware: Baremetal HP ProLiant DL360 Gen10

## Checklist

- [x] Documented in the inventory/hosts.sample
- [x] A change should not be merged unless it passes CI or there is a comment/update saying what testing was passed.

### DIFF
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -480,22 +480,42 @@
   - always
   - validation
 
-- name: "Set rendered_master_nmstate_yml from raw value for master nodes"
-  set_fact:
-    rendered_master_nmstate_yml: "{{ master_network_config_raw | to_yaml | trim }}"
-  when: master_network_config_raw is defined
+- name: Verify if master_network_config_template is defined and exists
+  stat:
+    path: "{{ master_network_config_template }}"
+    get_checksum: false
+  register: master_nm_template
+  when: master_network_config_template is defined
+  tags:
+  - always
+  - validation
 
-- name: "Set rendered_master_nmstate_yml from file for master nodes"
-  set_fact:
-    rendered_master_nmstate_yml: "{{ lookup('file', master_network_config_file) }}"
-  when: master_network_config_file is defined
+- name: Fail when master_network_config_template is defined but not exists
+  fail:
+    msg: "Variable master_network_config_template is defined but path does not exists"
+  when:
+    - master_network_config_template is defined
+    - not master_nm_template.exists|bool
+  tags:
+  - always
+  - validation
 
-- name: "Set rendered_worker_nmstate_yml from raw value for worker nodes"
-  set_fact:
-    rendered_worker_nmstate_yml: "{{ worker_network_config_raw | to_yaml | trim }}"
-  when: worker_network_config_raw is defined
+- name: Verify if worker_network_config_template is defined and exists
+  stat:
+    path: "{{ worker_network_config_template }}"
+    get_checksum: false
+  register: worker_nm_template
+  when: worker_network_config_template is defined
+  tags:
+  - always
+  - validation
 
-- name: "Set rendered_worker_nmstate_yml from file for worker nodes"
-  set_fact:
-    rendered_worker_nmstate_yml: "{{ lookup('file', worker_network_config_file) }}"
-  when: worker_network_config_file is defined
+- name: Fail when worker_network_config_template is defined but not exists
+  fail:
+    msg: "Variable worker_network_config_template is defined but path does not exists"
+  when:
+    - worker_network_config_template is defined
+    - not worker_nm_template.exists|bool
+  tags:
+  - always
+  - validation
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -492,7 +492,7 @@
 
 - name: "Set rendered_worker_nmstate_yml from raw value for worker nodes"
   set_fact:
-    rendered_master_nmstate_yml: "{{ worker_network_config_raw | to_yaml | trim }}"
+    rendered_worker_nmstate_yml: "{{ worker_network_config_raw | to_yaml | trim }}"
   when: worker_network_config_raw is defined
 
 - name: "Set rendered_worker_nmstate_yml from file for worker nodes"
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -495,7 +495,7 @@
     msg: "Variable master_network_config_template is defined but path does not exists"
   when:
     - master_network_config_template is defined
-    - not master_nm_template.exists|bool
+    - not master_nm_template.stat.exists|bool
   tags:
   - always
   - validation
@@ -515,7 +515,7 @@
     msg: "Variable worker_network_config_template is defined but path does not exists"
   when:
     - worker_network_config_template is defined
-    - not worker_nm_template.exists|bool
+    - not worker_nm_template.stat.exists|bool
   tags:
   - always
   - validation
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -497,5 +497,5 @@
 
 - name: "Set rendered_worker_nmstate_yml from file for worker nodes"
   set_fact:
-    rendered_master_nmstate_yml: "{{ lookup('file', worker_network_config_file) }}"
+    rendered_worker_nmstate_yml: "{{ lookup('file', worker_network_config_file) }}"
   when: worker_network_config_file is defined
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -479,3 +479,23 @@
   tags:
   - always
   - validation
+
+- name: "Set rendered_master_nmstate_yml from raw value for master nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ master_network_config_raw | to_yaml | trim }}"
+  when: master_network_config_raw is defined
+
+- name: "Set rendered_master_nmstate_yml from file for master nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ lookup('file', master_network_config_file) }}"
+  when: master_network_config_file is defined
+
+- name: "Set rendered_worker_nmstate_yml from raw value for worker nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ worker_network_config_raw | to_yaml | trim }}"
+  when: worker_network_config_raw is defined
+
+- name: "Set rendered_worker_nmstate_yml from file for worker nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ lookup('file', worker_network_config_file) }}"
+  when: worker_network_config_file is defined
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -485,6 +485,7 @@
     path: "{{ master_network_config_template }}"
     get_checksum: false
   register: master_nm_template
+  delegate_to: localhost
   when: master_network_config_template is defined
   tags:
   - always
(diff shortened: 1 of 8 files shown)


ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -479,3 +479,23 @@
...
+  when: master_network_config_raw is defined
+
+- name: "Set rendered_master_nmstate_yml from file for master nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ lookup('file', master_network_config_file) }}"
+  when: master_network_config_file is defined
+
+- name: "Set rendered_worker_nmstate_yml from raw value for worker nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ worker_network_config_raw | to_yaml | trim }}"
```
manurodriguez, line 495:
```suggestion
    rendered_worker_nmstate_yml: "{{ worker_network_config_raw | to_yaml | trim }}"
```

ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -479,3 +479,23 @@
...
+  when: worker_network_config_raw is defined
+
+- name: "Set rendered_worker_nmstate_yml from file for worker nodes"
+  set_fact:
+    rendered_master_nmstate_yml: "{{ lookup('file', worker_network_config_file) }}"
```
manurodriguez, line 500:
```suggestion
    rendered_worker_nmstate_yml: "{{ lookup('file', worker_network_config_file) }}"
```

manurodriguez (commented):
I noticed some typos, I'm using `rendered_master_nmstate_yml` facts twice, instead of  `rendered_worker_nmstate_yml`

manurodriguez:
From the last five tests, the one that failed was because I forgot to pass the file with the NMstate settings, but I retested and the succeeded jobs represent the following:

- [4.10](https://www.distributed-ci.io/jobs/5822a9f1-65c0-4deb-aa10-471c68f3f228/jobStates) validated NMstate settings are added in the install-config.yaml and nodes are deployed with such configuration.
- [4.11](https://www.distributed-ci.io/jobs/48d7d52f-1b81-4131-8625-544cf9a13c83/jobStates) validated NMstate settings are added in the install-config.yaml and nodes are deployed with such configuration.
- [4.9](https://www.distributed-ci.io/jobs/b7fc8973-b601-4a50-8bdd-626093cb7448/jobState) validated that even passing the variables the NMstate settings are not included in the install-config.yaml because the condition is only OCP >= 4.10
- [4.8](https://www.distributed-ci.io/jobs/dbb8599b-ee4b-46a6-aac1-4364b4dee6fc/jobStates) validated that deployments without NMstate settings continue working as expected. 

Ready for review!

manurodriguez:
@derekhiggins and @DirectedSoul1 I see by default you were placed as reviewers, could you please help us to review? Thanks!

derekhiggins:
> @derekhiggins and @DirectedSoul1 I see by default you were placed as reviewers, could you please help us to review? Thanks!

Hi @manurodriguez , this looks good to me, it results in a common networkConfig for nodes of the same role, I wonder if you have any thoughts on how we would expand this in the future to allow a networkconfig per node.

it doesn't have to be done in this PR I just want to ensure it will be possible in the future without breaking support for the API being created here.

manurodriguez:
> > @derekhiggins and @DirectedSoul1 I see by default you were placed as reviewers, could you please help us to review? Thanks!
> 
> Hi @manurodriguez , this looks good to me, it results in a common networkConfig for nodes of the same role, I wonder if you have any thoughts on how we would expand this in the future to allow a networkconfig per node.
> 
> it doesn't have to be done in this PR I just want to ensure it will be possible in the future without breaking support for the API being created here.

@derekhiggins Thanks for checking,

Initially I was debating between allowing networkconfig per node, or only one networkconfig for all nodes, so I decided to stay on the middle and do it for nodes of each role mainly because I couldn't visualize a clean way to do this per node.

My first option was to add networkconfig variable in the inventory per node, similar to the way we pass variables for disks, nics or other options, but because there are many `variable:value` options it was going to be a big json payload,  it looks ugly but is easy to transform that to yaml and add it in the install-config.yaml.j2 template per node. This is how the variables of a node would look like:

```
[masters]
master-0.cluster1.example.com name=master-0 ipmi_address=1.2.3.4 networkconfig='{"interfaces":[{"ipv4":{"auto-dns":true,"dhcp":true,"enabled":true},"ipv6":{"dhcp":false,"enabled":false},"link-aggregation":{"mode":"802.3ad","options":{"miimon":100,"mode":"802.3ad"},"port":["ens1f0","ens1f1"]},"mtu":9000,"name":"bond0","state":"up","type":"bond"},{"ipv4":{"auto-gateway":false,"auto-routes":false,"dhcp":true,"enabled":true},"ipv6":{"auto-gateway":false,"dhcp":false,"enabled":false},"mtu":9000,"name":"bond0.320","state":"up","type":"vlan","vlan":{"base-iface":"bond0","id":320}}]}' provision_mac=aa:bb:cc:dd:ee:ff ...
```
If we go with an option like this, I'll have to remove the tasks that take the variables: `master_network_config_file`, ` worker_network_config_file`, `master_network_config_raw`, and `worker_network_config_raw` but wouldn't take much changes in the install-config.yaml.j2 template.

My other option was to define only one main variable `networkconfig` and it will be an array to store more variables per node, for example:
```
networkconfig:
  - name: master-0
    interfaces:
    - name: <nic1_name>
      type: ethernet
      state: up
    ...
  - name: master-1
    interfaces:
    - name: <nic1_name>
      type: ethernet
      state: up
...
```
Then I'll have to iterate through the variable to look for node matches and include the variables in the install-config.yaml. This option will also require to remove the tasks that take the variables: `master_network_config_file`,  `worker_network_config_file`, `master_network_config_raw`, and `worker_network_config_raw`, and it would be only one task to load the variables from a file or raw, for example `network_config_file` and `network_config_raw` into `networkconfig` and the challenge will be in the install-config.yaml.j2 template to iterate through `networkconfig` and only include the variables if there is a node name match. 

In general I believe it's possible to expand this, but either way will require changes to a few tasks and variables I'm adding in this PR, unless I consider those changes now, if you have comments about any of the options above or even a different one please let me know, I can rework the change if is going to cause more troubles in the future, and allow passing the variables per node. At least now I have feedback, and we can go back ad forth until this looks good.

Thanks!

derekhiggins:
I agree option 1 above "add networkconfig variable in the inventory per node" would lead to a single line that is way too big

I like a combination of this PR (as it is) but somehow supporting an optional  per node section in the networkconfig but I agree to iterate through this in install-config.yaml.j2 would be a challenge 

One option that did occur to me would be to parse "rendered_master_nmstate_yml" with the perhost variable in scope
`{{ rendered_master_nmstate_yml | indent(10, true) | format(**hostvars[host]) }}`

this would allow us to add a static IP (this is the use case I had in mind) to the node variable in the inventory "e.g. staticip=1.2.3.4"

and then use it in the network config
```
     ipv4:
      address:
      - ip: "%(staticip)s"
        prefix-length: 24
      enabled: true

```
The main downside would be that "%" would become a special character, I'd prefer it to be jinja format {{}} but I could only figure out how to get python formatting to work , how would you feel about something like this?

manurodriguez:
I like your idea to keep using `rendered_master_nmstate_yml` and `rendered_worker_nmstate_yml` to define variables for all nodes of each role, and the optional per node selection. Also I think the use case to pass static IP address per node is valid and could work with this approach. 

Let me do some tests to visualize this better, and I'll get back with some ideas and hopefully not many questions :)

derekhiggins:
> I like your idea to keep using `rendered_master_nmstate_yml` and `rendered_worker_nmstate_yml` to define variables for all nodes of each role, and the optional per node selection. Also I think the use case to pass static IP address per node is valid and could work with this approach.
> 
> Let me do some tests to visualize this better, and I'll get back with some ideas and hopefully not many questions :)

thanks, btw if the optional per node selection thing gets messy I'd be happy with just the ability to set per node variables similar to the static ip above.

manurodriguez:
@derekhiggins we were testing a few options and I believe we found one method using only jinja format that allows to define specific network settings per node. Is very similar to what you suggested:

We are planning to use a `lookup` in the `templates/install-config.j2` for each master/worker role, for example:
```
{% if ((release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int >= 10)) %}
{% if master_network_config_template is defined and master_network_config_template|length %}
        networkConfig:
{{ lookup('template', master_network_config_template, template_vars=hostvars[host]) | indent(10, true) }}
{% endif %}
{% endif %}
```
The `master_network_config_template` is the file with the network settings in NMstate format as initially in this PR, but the user will have the liberty to prepare that file in jinja format, and any extra variables in the inventory can be used here to render the final configuration per node..

For example, assuming we have an inventory with these settings:
```
[masters]
master-0.cluster1.example.com name=master-0 ipmi_address=1.2.3.4 static_ip="9.9.8.8/24"
master-0.cluster1.example.com name=master-0 ipmi_address=1.2.3.4 static_ip="9.9.7.7/24"
master-0.cluster1.example.com name=master-0 ipmi_address=1.2.3.4 static_ip="9.9.6.6/24"
```
And  `master_network_config_template` file contains something like this:
```
interfaces:
- ipv4:
    enabled: true
    dhcp: false
    address:
    - ip: {{ static_ip.split('/')[0] }}
      prefix-length: {{ static_ip.split('/')[1] }}
```
Then the `lookup('template', master_network_config_template, template_vars=hostvars[host])` expression will pass any variables defined per host, and it's up to the user to define as many variables needed to render the template as desired.

Besides if the user does not want to pass any extra variables, and just wants to use the same static content for each node of the same role, then `master_network_config_template` can still be just a bunch of variables as illustrated in this PR, and the `lookup` will pretty much use anything in the template without rendering.

With this approach I'll remove the option to pass variables in raw, only `master_network_config_template`, `worker_network_config_template` will be used.

I'll update this PR and start deploying some clusters to validate, let me know if you have any comments.

manurodriguez:
@derekhiggins tests have been performed with latest changes:

- [4.11](https://www.distributed-ci.io/jobs/fa0507fd-11da-4903-bc81-78c106129b1c/jobStates?sort=date) validated NMstate settings are added in the install-config.yaml and nodes are deployed with such configuration.
- [4.10](https://www.distributed-ci.io/jobs/a9e26fcf-9c10-4fa0-975c-52a9489cbdf6/jobStates?sort=date) validated NMstate settings are added in the install-config.yaml and nodes are deployed with such configuration.
- [4.9](https://www.distributed-ci.io/jobs/bbab6f40-2e11-43df-aea2-965839216787/jobStates?sort=date) validated that even passing the variables the NMstate settings are not included in the install-config.yaml because the condition is only OCP >= 4.10
- [4.8](https://www.distributed-ci.io/jobs/e9f49bb4-0b05-4d04-a318-772a2b02c299/jobStates?sort=date) validated that deployments of OCP < 4.10 without NMstate settings continue working as expected. 
- [4.11*](https://www.distributed-ci.io/jobs/50b327c9-3401-470a-ba5f-51933d696cf9/jobStates?sort=date) validated that deployments of OCP >= 4.10 without NMstate settings continue working as expected. 

Ready again for review!

derekhiggins:
/lgtm
thanks looks good.
Looking at the failed unit tests, I think there are at least 2 problems, 
- A typo in the dockerfile name https://github.com/openshift/release/pull/34865
- A dockerfile wont build https://github.com/openshift-kni/baremetal-deploy/pull/927

derekhiggins:
Merging this, the fixes to the unit test haven't worked, will work on this in parallel

Issue #922, opened 2022-12-08
Permission issue when baremetal-deploy try to delete a /tmp/ansible.xxx folder


rh-gvincent:
baremetal-deploy downloads OpenShift binaries in a `/tmp/ansible.xxxx` folder using the tempfile ansible module. baremetal-deploy also remove all `/tmp/ansible.xxxx` folders before downloading those binaries. cf https://github.com/openshift-kni/baremetal-deploy/blob/47e20e26595fcf7696d7a6414cf4acb2af683a6e/ansible-ipi-install/roles/installer/tasks/10_get_oc.yml This is weird to me, but I may miss some context here.

The problem is baremetal-deploy is used by other ansible code that may use the tempfile ansible module with a user different than `kni`. The deletion of the folders will fail because of permission issue.

```
"rmtree failed: [Errno 13] Permission denied: '/tmp/ansible.lg268ohk'"
```

PR #923, opened 2022-12-08
Prefix temp directories with baremetal-deploy to avoid permission issues

rh-gvincent:
baremetal-deploy downloads OpenShift binaries in a `/tmp/ansible.xxxx` folder using the tempfile ansible module. baremetal-deploy also remove all `/tmp/ansible.xxxx` folders before downloading those binaries. cf https://github.com/openshift-kni/baremetal-deploy/blob/47e20e26595fcf7696d7a6414cf4acb2af683a6e/ansible-ipi-install/roles/installer/tasks/10_get_oc.yml This is weird to me, but I may miss some context here.

The problem is baremetal-deploy is used by other ansible code that may use the tempfile ansible module with a user different than `kni`. The deletion of the folders will fail because of permission issue.

This patch prefix the temp folder with baremetal-deploy. Binaries will be downloaded in `/tmp/baremetal-deploy.xxxx` temp folder.

Fixes: https://github.com/openshift-kni/baremetal-deploy/issues/922

### DIFF
--- a/ansible-ipi-install/roles/installer/tasks/10_get_oc.yml
+++ b/ansible-ipi-install/roles/installer/tasks/10_get_oc.yml
@@ -10,7 +10,7 @@
 - name: Find any old tmp dirs with OpenShift related binaries
   find:
     paths: /tmp
-    patterns: 'ansible.*'
+    patterns: "baremetal-deploy.*"
     file_type: directory
   register: tmp_results
   tags:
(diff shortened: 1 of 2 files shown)


rlmanrique:
+1 to this change.

In one of our setups we have issues because of this. I tested this patch and it fixes it, so it would be very helpful for us to have this merged.

rabin-io:
+1 as well, as we need to patch the code manually to make it work with our CI.

fredericlepied:
check dallas ocp-4.11-vanilla

fredericlepied (approved):

PR #974, opened 2023-07-10
Add externalMACAddress in virtualmedia template

### LINKED ISSUE #973: Virtual Media does not support externalMACAddress
In non-virtual media deployments the `externalMACAddress` is used for the external NIC on the bootstrap VM.

https://github.com/openshift-kni/baremetal-deploy/blob/ef54d0d1b0a3c6f8830ad04e8076afec3effb12d/ansible-ipi-install/roles/installer/templates/install-config.j2#L81-L83

But is not possible to do so when using virtual media, instead it is generated automatically by libvirt, it would be nice to allow virtual media deployments to also allow this setting.

tonyskapunk:
# Description

- `externalMACAddress` is lacking in the virtualmedia template, letting libvirt to generate the provisioner hwaddr, this change brings parity with non-virtualmedia installs that uses provisioning network
- Make `provisioningHostIP` required only in 4.6 where virtual media sets `provisioningNetwork: "Disabled"`

Fixes #973 

## Type of change

- [ x] New feature (non-breaking change which adds functionality)
- [ ] This change is a documentation update

## Testing

Tested in a lab with a disconnected environment.

**Test Configuration**:

- Versions: OCP 4.13
- Hardware: HP 

## Checklist

- [ ] A change should not be merged unless it passes CI or there is a comment/update saying what testing was passed.

### DIFF
--- a/ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
+++ b/ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
@@ -72,8 +72,13 @@ platform:
     externalBridge: {{ baremetal_bridge }}
 {% endif %}
     provisioningNetwork: "Disabled"
+{% if ((release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int == 6)) %}
     provisioningHostIP: {{ provisioningHostIP }}
+{% endif %}
     bootstrapProvisioningIP: {{ bootstrapProvisioningIP }}
+{% if externalMACAddress is defined and externalMACAddress|length %}
+    externalMACAddress: '{{ externalMACAddress }}'
+{% endif %}
 {% if bootstraposimage is defined and bootstraposimage|length %}
     bootstrapOSImage: {{ bootstraposimage }}
 {% endif %}
--- a/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
+++ b/ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
@@ -525,6 +525,7 @@
   when:
     - enable_virtualmedia|bool
     - provisioningHostIP is undefined
+    - release_version.split('.')[0]|int == 4 and release_version.split('.')[1]|int == 6
   tags:
     - always
     - validation


ansible-ipi-install/roles/node-prep/tasks/10_validation.yml
```diff
@@ -519,16 +519,6 @@
         msg: 'Adding hosts to dynamic HP group failed or redfish_inspection was set to false. All inventory systems will use IPMI.'
   tags: validation
 
-- name: Fail when provisioningHostIP and bootstrapProvisioningIP are not set when virtualmedia option is enabled
```
tonyskapunk, line 522: I've removed this validation because for newer versions this is not needed:

https://github.com/openshift-kni/baremetal-deploy/blob/ef54d0d1b0a3c6f8830ad04e8076afec3effb12d/documentation/ipi-install/modules/ipi-install-modifying-install-config-for-no-provisioning-network.adoc#L29-L36

I could re-add it with a conditional, but since 4.6 is already EOL I was wondering if it was worth doing it. 

https://github.com/openshift-kni/baremetal-deploy/blob/ef54d0d1b0a3c6f8830ad04e8076afec3effb12d/documentation/ipi-install/modules/ipi-install-modifying-install-config-for-no-provisioning-network.adoc#L11-L21

At this point the latest supported version is 4.10[1] , wdyt?

[1] https://access.redhat.com/support/policy/updates/openshift

tonyskapunk: In the end decided to keep the validation but only for the documented version (4.6) other versions do not require the `provisioningHostIP`

tonyskapunk:
At the moment we can't use the `check dallas ocp-4.13-vanilla` in here because our inventories need to set a couple of variables: `enable_virtualmedia` and   `bootstrapProvisioningIP` but in this job I used to test the changes I did it:

 https://www.distributed-ci.io/jobs/a658564c-8eba-4745-b14d-7881fb297be3/jobStates

cc: @derekhiggins

ansible-ipi-install/roles/installer/templates/install-config-virtualmedia.j2
```diff
@@ -72,8 +72,13 @@ platform:
     externalBridge: {{ baremetal_bridge }}
 {% endif %}
     provisioningNetwork: "Disabled"
+{% if ((release_version.split('.')[0]|int == 4) and (release_version.split('.')[1]|int == 6)) %}
     provisioningHostIP: {{ provisioningHostIP }}
```
derekhiggins, line 76: This change isn't mentioned in the description, does it need to be changed also?

tonyskapunk: @derekhiggins let me now if the description is better this way and if there is anything else to modify.

If needed I can trigger another test using an inventory with the expected vars, just let me know. Thanks!

PR #1064, opened 2024-06-20
Add not about maintenance status

derekhiggins:
This repository isn't being actively maintained, reflect that in the README. Also more clearly point out the location of where the ansible moved to.


derekhiggins:
cc @tonyskapunk any thoughts on doing this?

tonyskapunk:
I agree, it's better to be clear where the maintained code is found.

